Issue #2, opened 2021-02-18
Expand key_id to 8 bytes


enygren:
With key_id now being a hash of the public key, it would be preferable to switch the size of key_id to 8 bytes.  The chance of collisions with a  32-bit key_id with frequent rotations means that the key generation will need to track past keys to explicitly avoid collisions.  With a 64-bit key_id, accidental collisions is much more likely to be a non-issue.

enygren:
(We should also define the encoding of the public key being hashed)

chris-wood:
> We should also define the encoding of the public key being hashed

The public key is a DER-encoded SubjectPublicKeyInfo, which is fairly standard. 8 bytes seems fine here.

tfpauly:
I think this is done?

PR #3, opened 2021-02-18
Expand KeyID, and clarify its format.

### LINKED ISSUE #1: Allow privacy token to support either 2048 or 4096 RSA keys
We should support different versions here.

chris-wood:
Closes #1 and #2.

### DIFF
--- a/draft-privacy-token.md
+++ b/draft-privacy-token.md
@@ -19,6 +19,10 @@ author:
     city: Cupertino, California 95014
     country: United States of America
     email: tpauly@apple.com
+ -  ins: C. A. Wood
+    name: Christopher A. Wood
+    org: Cloudflare
+    email: caw@heapingbits.net
 
 
 normative:
@@ -58,9 +62,9 @@ of the structure contents.
 ~~~
 struct {
     uint8_t version;
-    uint8_t key_id[4];
+    uint8_t key_id[8];
     uint8_t message[32];
-    uint8_t signature[512];
+    uint8_t signature[Nk];
 } Token;
 ~~~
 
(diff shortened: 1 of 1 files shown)


draft-privacy-token.md
```diff
@@ -58,23 +62,25 @@ of the structure contents.
...
 } Token;
 ~~~
 
 The structure fields are defined as follows:
 
 - "version" is a 1-octet integer. This document defines version 1.
 
-- "key_id" is a 4-octet truncated key ID that identifies the key used to produce
-the signature. This is generated as SHA256(public key)[0:32].
+- "key_id" is a 8-octet truncated key ID that identifies the key used to produce
```
FredericJacobs, line 75: I thought we had agreed to 4 bytes?

FredericJacobs: Ooooh, I see. Now the request on the other issue.

8 bytes is still feasible to find a collision.

I'm discussing using the same blind signature scheme for another protocol where the KeyID forgery would break the security of the scheme.

Is there any opposition to just not truncate the hash?

FredericJacobs:
Random thought, if we require a hash function H for the blinding scheme, do we want to use that same hash function for the KeyID (eg. SHA384).

chris-wood:
> Random thought, if we require a hash function H for the blinding scheme, do we want to use that same hash function for the KeyID (eg. SHA384).

We could, yeah. But we might just consider changing things entirely so that the server specifies the KeyID and there is no client computation at all. (This is the direction we're going for ECH.) This is just a hint for servers anyway, and they know whether or not they will have collisions.

draft-privacy-token.md
chris-wood: None from me, but see my comment below where I propose the removing the need for clients to compute this at all.

plisiecki1:
8 or 16 bytes of hash should also reduce the odds of collisions to the extent that it should never happen. Letting the server pick the KeyID would also resolve the collision issue. There may be some (minor) convenience factor for servers to look up keys by a chosen KeyID (e.g., a counter) rather than a derived one (hash).

Rather than making Nk depend on the version, could the signature field for v1 be defined as the rest of the token?  Then no new version is needed for RSA2048 vs RSA4096 vs RSA8192, etc. As long as the client and proxy support the number of key bits, the signature can be encoded efficiently into a version 1 token.

chris-wood:
> Rather than making Nk depend on the version, could the signature field for v1 be defined as the rest of the token? Then no new version is needed for RSA2048 vs RSA4096 vs RSA8192, etc. As long as the client and proxy support the number of key bits, the signature can be encoded efficiently into a version 1 token.

Perhaps -- I'd need to think about this some more.

FredericJacobs:
@chris-wood : Do we just take a SHA-256 hash and not truncate it?

chris-wood:
> @chris-wood : Do we just take a SHA-256 hash and not truncate it?

What's the rationale? This is just a hint. If security of the thing using this hint depends on lack of collision, that seems like a bigger problem.

tfpauly (approved):

PR #6, opened 2021-03-23
Significant changes for Proxy DNS and SVCB request/response handling

enygren:
Significant changes and clarifications, renaming the existing headers and ending up with three headers:

    * Proxy-DNS-Request
    * Proxy-DNS-SVCB
    * Proxy-DNS-Used
   
Also specify client behavior to allow SVCB-optional clients to use proxies implementing these features while incorporating performance optimizations.

### DIFF
--- a/draft-proxied-svcb-headers.md
+++ b/draft-proxied-svcb-headers.md
@@ -398,10 +398,10 @@ specification to use SVCB records without performing additional DNS
 resolutions.  Clients doing so MUST implement other requirements
 specified in {{SVCB}} with the following providing a mechanism for
 doing so through a proxy.  As an example, if the client has a valid
-Proxy-DNS-SVCB header cached corresponding to a HTTPS RR, the client
-SHOULD upgrade to the "https" scheme as described in Section 3.5 of
-{{SVCB}}, which may involve abandoning a CONNECT to port 80
-through which it learned about the HTTPS RR.
+Proxy-DNS-SVCB header cached corresponding to a HTTPS RR (even if only
+for AliasMode), the client SHOULD upgrade to the "https" scheme as
+described in Section 3.5 of {{SVCB}}, which may involve abandoning a
+CONNECT to port 80 through which it learned about the HTTPS RR.
 
 If a client has a valid SVCB RRset or Proxy-DNS-SVCB cached for a
 given service, it SHOULD use CONNECT* with the authority hostname and
@@ -452,6 +452,21 @@ them as requests are made to the proxy.  Clients SHOULD periodically
 re-evaluate if new connections need to be established based on
 expiry of these TTLs.
 
+
+## SVCB-required clients
+
+SVCB-required clients must either perform additional SVCB
+DNS requests, or MUST use a proxy that is known and configured
+to always return Proxy-DNS-SVCB and Proxy-DNS-Used headers.
+
+SVCB-required clients SHOULD set high values for "wait" in
+Proxy-DNS-Request and MUST treat the absence of a Proxy-DNS-SVCB
+response header as an error.
+
+SVCB-required clients may also need to make their initial request to
+an authority with no expectation of being able to use that connection.
+Protocols specifying for SVCB-required clients will need to describe
+what clients should use in this case.
 
 
 # IANA Considerations
--- a/draft-proxied-svcb-headers.md
+++ b/draft-proxied-svcb-headers.md
@@ -1,4 +1,4 @@
-≈---
+---
 title: HTTP header fields for utilizing SVCB and HTTPS RRs via proxies
 abbrev: Proxied SVCB Headers
 docname: draft-proxied-svcb-headers-latest
@@ -78,7 +78,7 @@ for performing additional DNS lookups:
    by the proxy, allowing clients can proceed with the opportunistically
    established connection.  For services with SVCB records,
    the proxy will provide enough information to allow clients
-   to decide whether they can proceeed with using the connection
+   to decide whether they can proceed with using the connection
    or whether the client needs to establish a new connection
    through the proxy to the alternative endpoint specified
    in a SVCB record.  
@@ -179,7 +179,7 @@ HEADERS
 proxy-dns-request = "_foo.svc.example.com"; t=64; wait=400; params=(1 5); u; version=("draft-01")
 ~~~
 
-Additional examples are below in {{#examples}}.
+Additional examples are below in {#examples}.
 
 
 ## Proxy-DNS-SVCB Response Header Field {#proxy-dns-svcb}
@@ -188,7 +188,7 @@ A proxy server that receives a request with "Proxy-DNS-Request" MAY respond with
 the Structured Header "Proxy-DNS-SVCB" response header fields.
 
 The intent of this header is to provide SVCB-optional clients with
-enough information to implement {{SVCB} without performing additional
+enough information to implement {{SVCB}} without performing additional
 DNS lookups, including Sections 3 and 8.  This includes providing them
 with a list of alternative endpoints, as well as being explicit about
 whether SVCB records do and do not exist.
@@ -217,7 +217,7 @@ The TTL of the record MUST be a parameter with the key "ttl", and a value as an 
 This value must be the minimum TTL value of any CNAME or SVCB record encountered while
 resolving this SVCB record.
 
-SvcParams are represented with a parameter string contructed
+SvcParams are represented with a parameter string constructed
 prepending the string "key" to the numeric version of the
 SvcParamKey. For example, the ALPN SvcParamKey, with the numeric value
 1, would have a parameter key "key1". The value of each parameter MUST
@@ -364,18 +364,18 @@ HEADERS
 :method = CONNECT
 :status = 200
 proxy-dns-used = "svc.example.net.";ttl=7200;t=5;o="svc.example.com.",
-	         "svc2.example.net.";ttl=1800;t=5;o="svc.example.net.",
-	         "2001:db8::75";ttl=60;t=28;o="svc2.example.net."
+                 "svc2.example.net.";ttl=1800;t=5;o="svc.example.net.",
+                 "2001:db8::75";ttl=60;t=28;o="svc2.example.net."
 ~~~
 
 
 *TODO*: Do we need the "o=" parameter or is it redundant?
-	If we include it, should it be a MUST or MAY?
-	
+        If we include it, should it be a MUST or MAY?
+        
 *TODO*: Do we need to cover DNAME as well? 
 
 *TODO*: Should a future version be able to include NS record
-  	and DNSSEC information?
+        and DNSSEC information?
 
 
 # Proxy Behavior
@@ -457,9 +457,9 @@ re-evaluate if new connections need to be established based on
 expiry of these TTLs.
 
 *TODO*: Most of this section assumes ServiceMode records.
-	We should better structure to handle AliasMode records.
-	(I'm not sure the second condition ever applies?)
-	Below is some partial text:
+        We should better structure to handle AliasMode records.
+        (I'm not sure the second condition ever applies?)
+        Below is some partial text:
 
 When Proxy-DNS-SVCB (or a cached SVCB) contains an AliasMode record,
 clients SHOULD either use a connection made to the TargetName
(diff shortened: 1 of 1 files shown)


enygren:
Note that the build is broken due to some markdown issues:  I haven't been able to get even the original to build on my machine for some reason.

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,545 @@ author:
...
-DNS-SVCB-Keys = sf-list
+Proxy-DNS-Request = sf-item
 ~~~
 
-Each list member corresponds to the numeric version of an SvcParamKey.
+The item value is the name of a SVCB DNS query name.  This is the service
+name, possibly prepended with a label ({{SVCB}} Section 2.3).
+For example, "example.com" or "_foo.example.com" or "_8443._bar.api.example.com".
+
+*TODO*: Should this have a trailing "." to be an FQDN?
```
tfpauly, line 130: I'm a bit ambivalent here. I assume that the proxy can always append the "." when doing the lookup.

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,545 @@ author:
...
+
+The item has the following parameters:
+
+* t: The DNS RR type of the DNS query name, which MUST be specified
+  as an sf-integer (so 64 for SVCB and 65 for HTTPS).  If not
+  specified, the value of 65 (HTTPS) is the default.
+
+* wait: The maximum time the proxy should wait before responding to
+  the CONNECT* request while waiting for the SVCB resolution to
+  complete, specified as an sf-integer.  The proxy MAY choose to not
```
tfpauly, line 140: What's the unit here? Milliseconds?

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,545 @@ author:
...
+  If not specified, proxies SHOULD return the full set of SVCB parameters
+  for each SVCB RR.
+
+* version: Clients can optionally specify which versions of this specification
+  they support.  Its value MUST be an inner-list whose members are sf-string
+  values.  This current draft specifies version "draft-01".
 
-For example, a client wanting to receive ALPN and ECH Config parameters would
-send a request for 1 (alpn) and 5 (echconfig):
+* u: Clients can indicate whether they want DNS information about
```
tfpauly, line 158: Why `u`?

tfpauly: Ah, it's to get the Used response. Maybe call this out.

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,545 @@ author:
...
+(IPv4 or IPv6) that the proxy connected to for the CONNECT* request,
+or the value of a DNS CNAME on the path of aliases involved in
+resolving the hostname connected to by the proxy.
+
+The last member in the list MUST have an item value that is the string
+representation of the IPv6 or IPv4 address that represents the
+destination of the CONNECT* from the proxy.  If this is an IPv6
+address, it MUST follow the canonical string form from {{!RFC5952}}.
+
+*TODO*: Is there an IPv4 canonical form RFC we should reference here?
```
tfpauly, line 323: I don't think so? Closest I could find was https://tools.ietf.org/html/draft-main-ipaddr-text-rep-02

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,545 @@ author:
...
 :status = 200
-dns-svcb-params = "svc2.example.com.";priority=1;ttl=3600;p1=:aDIsaDM=:;p5=:MTIzLi4u:,
-                  "svc.example.com.";priority=2;ttl=3600;p1=:aDI=:;p5=:YWJjLi4u:
+proxy-dns-used = "svc.example.net.";ttl=7200;t=5;o="svc.example.com.",
+                 "svc2.example.net.";ttl=1800;t=5;o="svc.example.net.",
+                 "2001:db8::75";ttl=60;t=28;o="svc2.example.net."
 ~~~
 
+
+*TODO*: Do we need the "o=" parameter or is it redundant?
```
tfpauly, line 372: Not sure, I don't know if we'd use it on our client

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,545 @@ author:
...
+        If we include it, should it be a MUST or MAY?
+        
+*TODO*: Do we need to cover DNAME as well? 
+
+*TODO*: Should a future version be able to include NS record
```
tfpauly, line 377: Probably, yes... could that be added into this schema of types?

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,545 @@ author:
...
+        and DNSSEC information?
+
+
+# Proxy Behavior
+
+Proxy servers MUST NOT take action based on SVCB records.
+In particular, the ipv4hint and ipv6hint SvcParams MUST NOT
+be used by proxies for making connections.
+
+*TODO*: Discuss if this is too restrictive
```
tfpauly, line 387: Hm, why wouldn't they be able to do this? This does seem a bit restrictive.

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,545 @@ author:
...
+
+
+## Connection coalescing
+
+*TODO*: Do we want to include text on how clients
+can also use Proxy-DNS-Used for connection coalescing?
+(One thing missing there is the full set of IP addresses
+to look for overlaps within an address family.   I'm not
+sure this is worth the complexity.  Or do we remove this
+sub-section entirely?)
```
tfpauly, line 494: I certainly think clients should be free to use the information to determine coalescing policy, but I don't know if we need to detail it, or just can put a reference off to H2 text for that.

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,545 @@ author:
...
+Clients concerned about this may prefer to:
+* Not include "param" and receive all SVCB parameters.
+* Sort the item's parameters by their keys.
+* Limit variation in the values for "wait"
+  and select a wait value of 50, 100, 200,
+  400, 800, or 1600.
+* Not send "version" if implementing only the default
+  (published RFC) version.
+
+*TODO*: Is this a good set of "wait" values?
```
tfpauly, line 540: Do we need so many, I wonder? We could always just specify one good default to use.

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,545 @@ author:
...
 --- middle
 
 # Introduction
 
-CONNECT {{!RFC7231}} and CONNECT-UDP {{!I-D.ietf-masque-connect-udp}} are HTTP methods that
+CONNECT {{!RFC7231}} and CONNECT-UDP {{!I-D.ietf-masque-connect-udp}}
+are HTTP methods (referred hereafter as "CONNECT*") that
 clients may use to establish TCP or UDP flows to target servers. Once proxy servers establish
 these flows, proxy servers treat allocated flows as opaque byte or datagram streams respectively.
 Clients specify the target in authority-form (Section 5.3 of {{!RFC7230}}), including the name or
```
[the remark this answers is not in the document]
enygren: Perhaps we also add a Terminology section that gives names to a few of these things?  The "connect target" and "connect target host" are things we end up wanting to reference concretely in a bunch of places.

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,545 @@ author:
...
+  from DNS lookups in its response, including AAAA/A/CNAME
+  information used by the proxy in making its connection,
+  as well as SVCB records for a name specified in this header.
+
+This document also defines two new response headers:
+
+* Proxy-DNS-SVCB: encodes SVCB records associated with requested name.
+* Proxy-DNS-Used: encodes DNS information used by the Proxy for making this connection.
+
+*TODO*: do we need a Terminology section, or a reference to Section 1.4 of {{SVCB}} as well as {{!RFC8499}}?
```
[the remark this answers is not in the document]
enygren: Something like "connect target" and "connect target host" seem like good ones to include as mentioned above.  Perhaps it is a short section that only has one or two items for now but references Section 1.4 of {{SVCB}} for more Terminology?

draft-proxied-svcb-headers.md
enygren: I'm also ambivalent here.  We should pick one and specify, though.  Or specify that if the trailing "." is missing that the consumer should assume its presence and add it.  (The same applies for all of these containing DNS names.)

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,547 @@ author:
...
+
+*TODO*: Should this have a trailing "." to be an FQDN?
+
+The item has the following parameters:
+
+* t: The DNS RR type of the DNS query name, which MUST be specified
+  as an sf-integer (so 64 for SVCB and 65 for HTTPS).  If not
+  specified, the value of 65 (HTTPS) is the default.
+
+* wait: The maximum time the proxy should wait before responding to
```
enygren, line 143:
```suggestion
* wait: The maximum time in milliseconds the proxy should wait before responding to
```

draft-proxied-svcb-headers.md
enygren: Yes.  Added clarification.

draft-proxied-svcb-headers.md
enygren: Should we just call it "used"?  (especially since it is boolean and default-true so not included most of the time?)

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,550 @@ author:
...
+
+Each list member is an sf-string that represents either the IP address
+(IPv4 or IPv6) that the proxy connected to for the CONNECT* request,
+or the value of a DNS CNAME on the path of aliases involved in
+resolving the hostname connected to by the proxy.
+
+The last member in the list MUST have an item value that is the string
+representation of the IPv6 or IPv4 address that represents the
+destination of the CONNECT* from the proxy.  If this is an IPv6
+address, it MUST follow the canonical string form from {{!RFC5952}}.
```
enygren, line 326:
```suggestion
destination of the CONNECT* from the proxy.  If this is an IPv6
address, it MUST follow the canonical string form from {{!RFC5952}}.
If this is an IPv4 address, it MUST be in dotted octet form (four decimal
values with no leading zeros separated by ".").
```

draft-proxied-svcb-headers.md
enygren: Proposed some text above.  I couldn't find any either, and raising this 
to sunset4 doesn't seem like a good use of time.  ;)

draft-proxied-svcb-headers.md
enygren: Will need to think about this.  Perhaps we open an issue to think further on this after merging?

draft-proxied-svcb-headers.md
enygren: It's a topic for discussion.  Some reasons:
- This does change proxy semantics in normative ways.  (ie, it is changing how the proxy is deciding on what to connect to).  There may be security risks/exposures here, for example.  
- This could cause load-balancing issues as those may be smaller sets than A/AAAA and the TTLs may be longer.
This should probably get a first-class issue to discuss.  (I'm thinking a WG would also want to discuss this as a topic.)

draft-proxied-svcb-headers.md
enygren: Perhaps another issue-tracker item?  It's not core/critical here and we could drop or elaborate based on WG interest?

draft-proxied-svcb-headers.md
enygren: Maybe just pick 2-3?  Some will depend on the performance of the proxy to various DNS authorities.  SVCB-required clients may also may want a much higher value here than SVCB-optional.  I'd be happy reducing this set but we may need concrete data to propose a set.

enygren:
I'm confused --- this doesn't seem to be showing up in  https://github.com/tfpauly/privacy-proxy/blob/main/draft-proxied-svcb-headers.md ?

chris-wood:
Your other PR had commits from this PR, so GitHub probably got confused. You'll need to reopen this. :-(

PR #7, opened 2021-03-25
Support both rsa2048 and rsa4096

### LINKED ISSUE #1: Allow privacy token to support either 2048 or 4096 RSA keys

enygren:
Fixes #1 

Use version to differentiate (with v1=rsa4096 and v2=rsa2048).  There may be cleaner ways to address this,
but doing this retains wire compatibility with version 1.

### DIFF
--- a/draft-privacy-token.md
+++ b/draft-privacy-token.md
@@ -63,9 +63,10 @@ used with RSA Blind Signatures {{RSASIG}}.
 # Privacy Token Structure {#struct}
 
 A privacy token is a structure that begins with a single byte that indicates
-a version. This document defines version, 1, which indicates use of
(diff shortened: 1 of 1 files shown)


tfpauly:
@enygren I was chatting with @chris-wood , and we’re thinking of just relaxing the version to not for the Nk length, actually. Do you see any issue with that?

enygren:
So allowing Nk to be variable and the remainder of the token size for v1?  That would also be fine/better.

chris-wood:
Thanks, @enygren :-)

enygren:
Abandoning in-favor of an alternate PR, coming shortly.

PR #9, opened 2021-03-30
Significant changes for Proxy DNS and SVCB request/response handling : try 2

enygren:


enygren:
Hopefully this should actually be able to be merged?

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,551 @@ author:
...
+  "NXDOMAIN" response when resolving the SVCB DNS query name specified
+  in the Proxy-DNS-Request header), the list MUST contain a single
+  entry with the item value of ".", indicating that the name has no
+  SVCB records.  This special value of "." MUST NOT be returned in
+  error conditions such as timeouts.  The value of the "ttl" parameter
+  MUST be the TTL of this authoritative DNS response.
+
+* If an error condition is encountered (such as a timeout, loop, or
+  invalid record), the Proxy-SVCB-DNS header MUST contain an empty list
+  (and thus not be returned).
```
MikeBishop, line 273: I'm not sure these are the same thing.  How about simply....
```suggestion
  invalid record), the Proxy-SVCB-DNS header MUST NOT be returned.
```

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,551 @@ author:
...
+
+A successful CONNECT response would include the following headers, if the client requested both
+"alpn" and "echconfig":
+
+~~~ example
+HEADERS
+:method = CONNECT
+:status = 200
+proxy-dns-svcb = "svc2.example.net.";priority=1;ttl=1800;key1=:aDIsaDM=:;key5=:MTIzLi4u:,
+                 "svc.example.net.";priority=2;ttl=1800;key1=:aDI=:;key5=:YWJjLi4u:
```
MikeBishop, line 296: Given that the ellipsis is intended to represent longer content in the field, it might be better not to encode literal `...` in these binary examples.
```suggestion
proxy-dns-svcb = "svc2.example.net.";priority=1;ttl=1800;key1=:aDIsaDM=:;key5=:MTIz...:,
                 "svc.example.net.";priority=2;ttl=1800;key1=:aDI=:;key5=:YWJj...:
```

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,551 @@ author:
...
+(IPv4 or IPv6) that the proxy connected to for the CONNECT* request,
+or the value of a DNS CNAME on the path of aliases involved in
+resolving the hostname connected to by the proxy.
+
+The last member in the list MUST have an item value that is the string
+representation of the IPv6 or IPv4 address that represents the
+destination of the CONNECT* from the proxy.  If this is an IPv6
+address, it MUST follow the canonical string form from {{!RFC5952}}.
+
+*TODO*: Is there an IPv4 canonical form RFC we should reference here?
```
MikeBishop, line 328: We might use `host` out of RFC 3986, since it specifies IPv4, IPv6, and hostname rendering.

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,551 @@ author:
...
+
+If the hostname connected to by the proxy was a series of CNAMEs,
```
MikeBishop, line 330: I understand the point, but I don't know that a hostname can be said to "be" a series of CNAMEs.  Perhaps...
```suggestion
If the hostname provided by the client referred to one or more
CNAMEs before resolving to an IP address,
```

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,551 @@ author:
...
+  as an sf-integer (so 1 for A, 28 for AAAA, and 5 for CNAME).
+
+* ttl: The TTL of the address or CNAME record SHOULD be included as a
+  parameter with the key "ttl", whose value MUST be an sf-integer.
+  If the value of of intermediate records are unavailable,
+  the ttl parameter SHOULD be included on just the last list
+  entry, containing the minimum TTL value across the CNAMEs.
+
+* o: The owner name of the DNS record MAY be included.
+  This this parameter is specified, its value MUST be an sf-string
```
MikeBishop, line 346:
```suggestion
  If this parameter is specified, its value MUST be an sf-string
```

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,551 @@ author:
...
 ~~~
 
+
+*TODO*: Do we need the "o=" parameter or is it redundant?
+        If we include it, should it be a MUST or MAY?
+        
+*TODO*: Do we need to cover DNAME as well? 
+
+*TODO*: Should a future version be able to include NS record
+        and DNSSEC information?
```
MikeBishop, line 383: At first glance, it appears to be redundant. Is there a situation in which the owner is not simply the original hostname (for the first entry) or the preceding hostname (for subsequent entries)?

However, if we want to be able to include additional information in the future, it might make sense to swap the format around; rather than including a list of CNAME(s) and/or IP(s) and giving an owner for each, list the owner first and an inner list of useful records for each.

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,551 @@ author:
...
+
+
+# Proxy Behavior
+
+Proxy servers MUST NOT take action based on SVCB records.
+In particular, the ipv4hint and ipv6hint SvcParams MUST NOT
+be used by proxies for making connections.
+
+*TODO*: Discuss if this is too restrictive
```
MikeBishop, line 392: I think this probably is too restrictive, especially in the case of SVCB-naive clients.  It means a high probability of the client having to open a new tunnel after processing the records, even though avoiding that is something this spec states as a goal.

It still seems reasonable for the proxy to chase AliasMode records, use hints, etc. to choose the target to connect to.  The flip side, I suppose, would be if the proxy chooses a host whose ALPN list doesn't reflect the client's support, but that could be solved by requiring the server to choose based on the transport protocol implied by the method, and then preferring the default ALPN within that.  Alternatively, the client could include an ALPN list in the request.

However, it's also totally reasonable to save that for a separate PR if we decide to do it.

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,551 @@ author:
...
+
+
+# Client Behavior
+
+Clients that are SVCB-required ({{SVCB}} Section 3) MUST perform a DNS
+resolution prior to making a CONNECT* request, as they will need to
+obtain SVCB records and a TargetName.
```
MikeBishop, line 399: Given the problems of doing such a resolution noted above, how would they do that?  It might make more sense to say that a client which is SVCB-required must abandon the tunnel if it doesn't receive SVCB records from the proxy.

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,551 @@ author:
...
+
+
+## Connection coalescing
+
+*TODO*: Do we want to include text on how clients
+can also use Proxy-DNS-Used for connection coalescing?
+(One thing missing there is the full set of IP addresses
+to look for overlaps within an address family.   I'm not
+sure this is worth the complexity.  Or do we remove this
+sub-section entirely?)
```
MikeBishop, line 499: My vote is for removal.  Coalescing based on IP address is not a battle we've been winning, and I don't think this draft is the venue to continue fighting it.

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,551 @@ author:
...
+
+
 # IANA Considerations
 
 ## HTTP Headers {#iana-header}
 
-This document registers the "DNS-SVCB-Keys" and "DNS-SVCB-Params",
+This document registers the "Proxy-DNS-Request" and "Proxy-DNS-SVCB",
 headers in the "Permanent Message Header Field Names"
 <[](https://www.iana.org/assignments/message-headers)>.
```
MikeBishop, line 519: HTTP-core is moving these registries and will likely be an RFC before this document.  Let's go ahead and point there.

MikeBishop (commented):
This is a good start, with mostly nits and comments on issues that can be deferred from this PR.

I do feel like we don't have a good story (or perhaps merely a good example?) for cases which involve a mix of CNAMEs and AliasMode entries, though perhaps that's alleviated by the current design of telling the proxy not to consider SVCB records in performing the actual connection.

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,551 @@ author:
...
+  only include SVCB information it has cached.
+  If this parameter is not specified, proxies are free to choose
+  a reasonable default.
+
+* params: Clients can optionally limit which SVCB parameters they'd like
+  to receive by optionally specifying this parameter.  Its value MUST
+  be an inner-list whose members are sf-integer items.
+  Each list member corresponds to the numeric version of an SvcParamKey.
+  If not specified, proxies SHOULD return the full set of SVCB parameters
+  for each SVCB RR.
```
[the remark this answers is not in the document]
enygren: Just strips out other parameters.  Is this not clear later?

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,551 @@ author:
...
+
+* version: Clients can optionally specify which versions of this specification
+  they support.  Its value MUST be an inner-list whose members are sf-string
+  values.  This current draft specifies version "draft-01".
```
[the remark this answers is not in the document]
enygren: Just when breaking changes.

draft-proxied-svcb-headers.md
enygren: In structured headers they are the same thing.  Would saying an empty sf-list be more clear?

draft-proxied-svcb-headers.md
enygren, line 330:
```suggestion
If the hostname connected to by the proxy aliased to one or more CNAMEs,
```

enygren: (incorporating this but with "aliased" rather than "referred"

draft-proxied-svcb-headers.md
enygren: The redundancy is why I'm include to leave it optional or just get rid of it.  It could be helpful for cases such as DNSSEC in the future, or perhaps if we include NS records or other things,

draft-proxied-svcb-headers.md
enygren: Part of why I'm wary of this is that it changes the Proxy behavior and semantics substantially.  Rather than the proxy just returning additional information, here we start fundamentally changing behavior.  (Would doing this count as an "Update" to any of the RFCs defining CONNECT, for example?)

draft-proxied-svcb-headers.md
enygren: They could do DoH through the proxy, for example.  Or use ODoH.  Or do something else.  We could mention this in security considerations, but it seems like an implementation detail we don't want to get into?  This is already covered in the core SVCB draft in some detail.

draft-proxied-svcb-headers.md
enygren, line 499:
```suggestion

```

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,551 @@ author:
...
+*TODO*: Write this section...
+
+
+
 # IANA Considerations
 
 ## HTTP Headers {#iana-header}
 
-This document registers the "DNS-SVCB-Keys" and "DNS-SVCB-Params",
+This document registers the "Proxy-DNS-Request" and "Proxy-DNS-SVCB",
```
enygren, line 517:
```suggestion
*TODO*: Move the registries below to the new HTTP-core registry.

This document registers the "Proxy-DNS-Request" and "Proxy-DNS-SVCB",
```

draft-proxied-svcb-headers.md
enygren: Added a todo

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,548 @@ author:
...
+being DNSSEC-validated MUST NOT use this metadata without otherwise
+fetching the record and its corresponding RRSIG record and locally
+verifying its contents.
+
+Clients relying on ECH should avoid sending anything on
+opportunistically created connections until verifying whether
+there is a preferred alternative service that supports ECH
+which they should use.
+
+
```
enygren, line 560:
```suggestion

# Acknowledgments

Thank you to Mike Bishop, Chris Wood, and others for their feedback and suggestions on this draft.
```

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,551 @@ author:
...
+# Acknowledgments
+
+Thank you to Mike Bishop, Chris Wood, and others for their feedback and suggestions on this draft.
+# Appendix: Additional Examples {#examples}
+
```
enygren, line 565:
```suggestion
Thank you to Mike Bishop, Chris Wood, 
and others for their feedback and suggestions on this draft.

# Appendix: Additional Examples {#examples}

```

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,551 @@ author:
...
+  be an inner-list whose members are sf-integer items.
+  Each list member corresponds to the numeric version of an SvcParamKey.
+  If not specified, proxies SHOULD return the full set of SVCB parameters
+  for each SVCB RR.
+
+* version: Clients can optionally specify which versions of this specification
+  they support.  Its value MUST be an inner-list whose members are sf-string
+  values.  This current draft specifies version "draft-01".
+
+* u: Clients can indicate whether they want DNS information about
```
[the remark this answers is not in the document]
enygren: yes, A/AAAA/CNAME records
```suggestion
* u: Clients can indicate whether they want DNS information (A, AAAA,
    and CNAME records) about
```

Smithad499:
> 

eaffef1522713c529e12b9f281470cba27768b23

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,551 @@ author:
...
-This document specifies HTTP header fields that proxy servers may use to relay information retrieved
-from SVCB records from proxy servers to clients when using CONNECT or CONNECT-UDP.
+It is not always optimal for client that want to use SVCB to perform a separate
+DNS resolution prior to using a CONNECT* proxy, for a couple reasons:
+
+1. The extra DNS lookup incurs a performance penalty in delaying the client's
+connection establishment, which might be wasted if there aren't any SVCB records present.
+
+2. If the client is using the proxy for providing additional privacy, performing
+DNS lookups not through the proxy might disclose the client's destination
```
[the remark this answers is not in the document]
Smithad499: Great job.

draft-proxied-svcb-headers.md
```diff
@@ -17,132 +17,551 @@ author:
...
-a connection to the target.
+proxy server locally resolves the name to an IPv4 or IPv6 address with A or AAAA queries. As previously
+defined, clients do not see these A or AAAA answers.
+
+*TODO*: some CONNECT* methods may be moving away from specifying the target
+in authority-form (and may use something closer to what is specified in rfc8441)
+and we should update accordingly.  
+
+Clients using Service Binding (SVCB or HTTPS) records {{!SVCB=I-D.ietf-dnsop-svcb-https}}
+need perform additional DNS resolutions prior to issuing a CONNECT* request, as clients
```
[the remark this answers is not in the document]
Smithad499: It's DRE

PR #29, opened 2021-09-15
Make PATs interactive.

### LINKED ISSUE #20: Private Access Tokens should have an interactive mode
Instead of being able to use cached tokens, there should also be a way to have the origin have a one-time challenge that binds a token to a particular session. This is important for high-value tokens that shouldn't be possible to harvest.

chris-wood:
Closes #20, making interactive the only mode.

### DIFF
--- a/draft-private-access-tokens.md
+++ b/draft-private-access-tokens.md
@@ -90,12 +90,15 @@ The Issuer is able to see the identity of the Redeemer (ORIGIN_ID), but only see
 anonymous identifier for a client (ANON_CLIENT_ID). Issuers maintain the details of
 policy enforcement on behalf of the Redeemer. For example, a given policy might be,
 "issue at most N tokens to each client." Example policies and their use cases are
-discussed in {{policies}}.
+discussed in {{examples}}.
 
 The Redeemer, which represents the service being accessed by the client, only receives
 a Private Access Token from the client.
 
-## Terminology {#terms}
+## Notation and Terminology {#terms}
+
+Unless said otherwise, this document encodes protocol messages in TLS notation
+from {{!TLS13=RFC8446}}, Section 3.
 
 The following terms are defined to refer to the different pieces of information
 passed through the system:
@@ -110,6 +113,12 @@ ISSUER_POLICY_WINDOW:
 policy, defined in terms of seconds and represented as a uint64. The ANON_CLIENT_ID
 that the Mediator derives is specific to a Policy Window, meaning that a CLIENT_ID
 will not map to the same ANON_CLIENT_ID after the Policy Window has elapsed.
+
+ISSUER_KEY:
+: The public key used when generating and verifying Private Access Tokens.
+
+ORIGIN_NAME:
+: The Origin name identifies the Redeemer that requests Private Access Tokens.
 
 ORIGIN_ID:
 : The Origin Identifier represents the service for which the Client is requesting a
@@ -139,22 +148,25 @@ and corresponds to a unique ANON_ORIGIN_ID and a unique ANON_CLIENT_ID.
 It is assumed that Issuers make Oblivious HTTP configurations and policy verification
 keys available via the following API endpoints:
 
-- OHTTP configuration: /.well-known/ohttp-config
+- OHTTP configuration: /.well-known/ohttp-configs
 - Access token policy verification key: /.well-known/access-token-key/policy=?
 - Access token policy window: /.well-known/access-token-window/policy=?
 
 The OHTTP configuration is defined in {{!OHTTP=I-D.thomson-http-oblivious-http}}.
+The response is a serialied collection of configurations, and uses media
+type "application/ohttp-keys".
+
 The access token policy verification key is a struct of the following format:
 
 ~~~
 struct {
-  opaque public_key[Nk]; // Defined in [BLINDSIG]
+   opaque public_key[Nk]; // Defined in [BLINDSIG]
 } AccessTokenKey;
 ~~~
 
 The contents of AccessTokenKey are an RSA public key for use with the RSA Blind
 Signature protocol {{!BLINDSIG=I-D.irtf-cfrg-rsa-blind-signatures}}. The response
-use media type "application/access-token-key".
+uses media type "application/access-token-key".
 
 The access token policy window is a resource of media type "application/json", with the
 following structure:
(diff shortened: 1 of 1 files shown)


draft-private-access-tokens.md
```diff
@@ -173,14 +185,83 @@ Mediators advertise an Oblivious HTTP URI template {{!RFC6570}} for proxying
...
+: Name of the Redeemer (ORIGIN_NAME).
+
+issuer_name:
+: Name of the Issuer (ISSUER_NAME).
+
+redemption_nonce:
+: A fresh 32-byte nonce generated for each redemption request.
+
+timestamp:
+: The current UNIX epoch in seconds.
```
dvorak42, line 233: nit: "current UNIX epoch time"

draft-private-access-tokens.md
```diff
@@ -173,14 +185,83 @@ Mediators advertise an Oblivious HTTP URI template {{!RFC6570}} for proxying
 protocol messages to Issuers. For example, one template
 for the Mediator might be https://mediator.net/relay-access-token-request.
 
-# Issuance
+# Protocol
```
dvorak42, line 188: Do we want to only have a purely interactive protocol?

While interactivity does improve some of the security properties through immediate redemption binding, it still seems nice to have a non-interactive protocol to avoid high cost latency issues, or even requiring the mediator to have to store state for non-authenticated clients.

chris-wood: I'm not sure the flexibility is warranted here, but maybe we could support both.

draft-private-access-tokens.md
```diff
@@ -173,14 +185,83 @@ Mediators advertise an Oblivious HTTP URI template {{!RFC6570}} for proxying
...
+Clients then present this token to Redeemers using the Authorization header as follows:
 
-Issuance assumes the Client has the following information:
+~~~
+Authorization: PrivacyToken t=abc
+~~~
+
+Where the token is a serialized Private Access Token corresponding to the given Redeemer
+policy, and the PrivacyToken message is SHA256(context). Redeemers verify the token using
+the corresponding policy verification key from the Issuer.
```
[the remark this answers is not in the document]
chris-wood: Presumably it would be bound to the session, or clients could echo back the context when presenting a token?

draft-private-access-tokens.md
```diff
@@ -173,14 +185,83 @@ Mediators advertise an Oblivious HTTP URI template {{!RFC6570}} for proxying
...
+redemption_nonce:
+: A fresh 32-byte nonce generated for each redemption request.
+
+timestamp:
+: The current UNIX epoch in seconds.
+
+This context is sent to Clients in an "WWW-Authenticate" header as follows:
+
+~~~
+WWW-Authenticate: PrivacyToken realm="<context>"
```
tfpauly, line 238: I'm not sure the context should be stuck in the "realm" variable. Can we have a new variable here? And we should likely clarify it needs to be a quoted base64 string.

draft-private-access-tokens.md
```diff
@@ -173,14 +185,83 @@ Mediators advertise an Oblivious HTTP URI template {{!RFC6570}} for proxying
...
+~~~
+
+Upon receipt, Clients use the context in the Issuance protocol, described in {{issuance}}.
+The output of that protocol is a PrivacyToken bound to the redemption context.
+Clients then present this token to Redeemers using the Authorization header as follows:
 
-Issuance assumes the Client has the following information:
+~~~
+Authorization: PrivacyToken t=abc
```
tfpauly, line 246:
```suggestion
Authorization: PrivacyToken token=abc
```

draft-private-access-tokens.md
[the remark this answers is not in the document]
tfpauly: With the interactive mode, I think it would need to be tied to the HTTP session, right?

draft-private-access-tokens.md
```diff
@@ -173,14 +185,83 @@ Mediators advertise an Oblivious HTTP URI template {{!RFC6570}} for proxying
...
+a Private Access Token. See {{access-token-keys}} for details on how this
+can be obtained.
+
+Token redemption is an interactive protocol. Redeemers challenge Clients to
+present a unique, single-use token. Redeemers present this challenge to Clients
+with the following context:
+
+~~~
+struct {
+   uint8 policy_window;
```
tfpauly, line 212: Why is the policy window needed here?

draft-private-access-tokens.md
```diff
@@ -173,14 +185,83 @@ Mediators advertise an Oblivious HTTP URI template {{!RFC6570}} for proxying
...
+   opaque redeemer_name<1..2^16-1>;
```
tfpauly, line 213: If this is bound the origin, this may be redundant—or we have it to ensure that the values match.

draft-private-access-tokens.md
dvorak42: Yeah, agreed that the interactive one is cleaner/simpler. Maybe just have the interactive mode for now, and maybe add a note somewhere that a non-interactive construction is possible in case an infra/use-case warrants it?

draft-private-access-tokens.md
dvorak42: Maybe just call it "challenge" or "context". Though if the redemption origin is in the context, we don't need a separate realm?

draft-private-access-tokens.md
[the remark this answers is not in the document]
dvorak42: Yeah, I don't think the client could be echoing back, otherwise it would lie about the previous nonce (unless you have some sort of signed bit there).

draft-private-access-tokens.md
tfpauly: I'd like to copy this: https://www.rfc-editor.org/rfc/rfc7486.html#section-3

draft-private-access-tokens.md
```diff
@@ -173,14 +185,83 @@ Mediators advertise an Oblivious HTTP URI template {{!RFC6570}} for proxying
...
+   opaque issuer_name<1..2^16-1>;
+   opaque redemption_nonce[32];
+   uint64 timestamp;
+} RedemptionContext;
```
tfpauly, line 217: Call this a challenge, not a context?

draft-private-access-tokens.md
tfpauly: Let's remove this from the challenge struct for now

draft-private-access-tokens.md
```diff
@@ -173,14 +185,83 @@ Mediators advertise an Oblivious HTTP URI template {{!RFC6570}} for proxying
...
+present a unique, single-use token. Redeemers present this challenge to Clients
+with the following context:
+
+~~~
+struct {
+   uint8 policy_window;
+   opaque redeemer_name<1..2^16-1>;
+   opaque issuer_name<1..2^16-1>;
+   opaque redemption_nonce[32];
+   uint64 timestamp;
```
tfpauly, line 216: I think we can remove the timestamp here, for simplicity

draft-private-access-tokens.md
chris-wood: Yeah, "challenge" is good!

tfpauly (approved):

PR #40, opened 2021-09-20
Add single issuer-mediator considerations

smhendrickson:
Feels a bit wordy. @dvorak42, is there a specific link for the privacy promises of privacy pass?

Should we just use https://datatracker.ietf.org/doc/draft-ietf-privacypass-protocol/ ?

### DIFF
--- a/draft-private-access-tokens.md
+++ b/draft-private-access-tokens.md
@@ -548,6 +548,17 @@ Clients require the access token verification key before requesting a token. Cli
 NOT connect directly to the issuer to fetch any such keys. Instead, clients SHOULD use an
 anonymizing request proxy such as OHTTP {{OHTTP}}.
 
+# Privacy Considerations {#privacy-considerations}
+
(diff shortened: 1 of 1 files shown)


draft-private-access-tokens.md
```diff
@@ -548,6 +548,17 @@ Clients require the access token verification key before requesting a token. Cli
 NOT connect directly to the issuer to fetch any such keys. Instead, clients SHOULD use an
 anonymizing request proxy such as OHTTP {{OHTTP}}.
 
+# Privacy Considerations {#privacy-considerations}
+
+## Issuer and Mediator Ownership
+
+Issuers and Mediators should be run by mutually distinct organizaitons to limit
```
tfpauly, line 555:
```suggestion
Issuers and Mediators should be run by mutually distinct organizations to limit
```

tfpauly (approved):

Issue #34, opened 2021-09-16
Describe captcha/fraud rate limiting use case alongside metered paywalls


tfpauly:

chris-wood:
@tfpauly still planning on something here?

tfpauly:
I think it's covered adequately now

PR #42, opened 2021-09-22
Add motivating use cases

janaiyengar:


draft-private-access-tokens.md
```diff
@@ -80,10 +80,79 @@ separation between the parties involved in the issuance and the redemption
...
+A specific example of this use case is a metered paywall, where an origin limits
+the number of page requests to each unique user over a period of time before the
+user is required to pay for access. The origin typically resets this state
+periodically, say, once per month. For example, an origin may serve ten (major
+content) requests in a month before a paywall is enacted.
+
+Origins routinely use client IP addresses for this purpose.
+
+
+## Client Geo-Location
```
tfpauly, line 113: We were discussing this case a bit with @dvorak42 and @smhendrickson; as long as Issuers are willing to maintain a key per geo (as we do in Private Relay), this only needs to change the mechanism by having a new property that the Mediator can check from the Issuer to understand the geo-policy to enforce per key.

There was a discussion of if this should a mechanism that's described in this document, or a small extension document that focuses on the geolocation use case specifically. Thoughts?

draft-private-access-tokens.md
```diff
@@ -80,10 +80,79 @@ separation between the parties involved in the issuance and the redemption
...
+
 ## Requirements
 
 {::boilerplate bcp14}
 
+
+# Motivation
+
+This section describes classes of use cases where an origin would traditionally
+use a stable unique client identifier for enforcing policy. Hiding these
```
chris-wood, line 92:
```suggestion
use a stable and unique client identifier for enforcing attribute-based policy. Hiding these
```

Attributes are listed below in the use cases and include geo/region, service account existence, etc.

chris-wood (commented):
These use cases are very clearly written -- nice! I wonder if we should de-dupe the now redundant text in the example policy section?

draft-private-access-tokens.md
smhendrickson: As discussed we didn't feel the geolocation case scales well beyond country, due to the number of keys required. We felt public metadata was a stronger approach, but requires different crypto to be standardized first.

I think it is ok to leave this example because using different keysets should be supported as an interim approach. We should however assume any of these examples will need a bit of extension standards to operate smoothly. Geo is a great place to start.

draft-private-access-tokens.md
```diff
@@ -80,10 +80,79 @@ separation between the parties involved in the issuance and the redemption
...
+for addressing these use cases is described in {{examples}}.
+
+## Rate-limited Access
+
+An origin provides rate-limited access to content to a client over a fixed
+period of time. The origin does not need to know the client's identity, but
+needs to know that a requesting client has not exceeded the maximum rate set by
+the origin.
+
+A specific example of this use case is a metered paywall, where an origin limits
```
tfpauly, line 104:
```suggestion
One example of this use case is a metered paywall, where an origin limits
```

draft-private-access-tokens.md
```diff
@@ -80,10 +80,79 @@ separation between the parties involved in the issuance and the redemption
...
+the number of page requests to each unique user over a period of time before the
+user is required to pay for access. The origin typically resets this state
+periodically, say, once per month. For example, an origin may serve ten (major
+content) requests in a month before a paywall is enacted.
+
```
tfpauly, line 109:
```suggestion
Another example of rate-limited access is in fraud-prevention mechanisms. Operations that 
are sensitive to fraudulent abuse, like using a website to create an account, often want to
rate-limit access, and start displaying captchas or require extra verification if a rate-limit
is exceeded.

```

draft-private-access-tokens.md
```diff
@@ -80,10 +80,79 @@ separation between the parties involved in the issuance and the redemption
...
+Origins routinely use client IP addresses for this purpose.
```
tfpauly, line 110:
```suggestion
Origins routinely use client IP addresses for this purpose. The per-client policy enforcement
possible with Private Access Tokens allows for rate-limiting on the level of a client device or
user account, instead of client IP address.
```

tfpauly (commented):
Adding some suggestions to bring in the captcha case (which would address #34). I also think we can remove the "Policies and Uses Cases" section with this change.

draft-private-access-tokens.md
janaiyengar: I don't think you should add this text here. This belongs in {{examples}}, where the solution using PATs is described.

draft-private-access-tokens.md
janaiyengar: I am not sure I follow (and this could well be because I missed the discussion and/or I am missing something more fundamental). I think I'm confused in part by @smhendrickson 's statement of "ok to leave this example", which parses ambiguously in my mind. Did you mean "leave it in" or "leave it out"?

Let me share my thoughts on two things anyway:

- I think this depends in part on what this document is outlining. I think it outlines an architecture as well as a protocol. Given it does both, I would argue that the architecture is employed here for _all_ of these use cases.

- I agree that the solution here is going to be limited to per-country or per-region and not arbitrary geo-location. I'm not sure I understand the proposal though, since at the moment the only way we can do this now is with key per region, and I agree that this won't scale beyond per country-ish numbers. However, you can either do it or not... and we don't have the other necessary crypto bits yet. I would love to work on that separately (and understand where the crypto required for that stands), but per-country geo locating, while a bit limited, is a pretty important use case.

draft-private-access-tokens.md
tfpauly: Fair, that works

draft-private-access-tokens.md
tfpauly: @janaiyengar my impression is that @smhendrickson was saying to "leave it in", and keep the example.

tfpauly (approved):

draft-private-access-tokens.md
janaiyengar: Apologies for being dense, @smhendrickson :-)

smhendrickson: Sorry just seeing this. Yup, I meant leave it in because we don't yet have the crypto to support something fancier.

@janaiyengar no worries :)

PR #45, opened 2021-09-23
Expand token security considerations

### LINKED ISSUE #14: Clarify that Private Access Tokens are bound to origins
We should be clear that token keys are per-origin, and not used for different origins within a site

### LINKED ISSUE #19: Discuss TLS MITM for Private Access Tokens
To add to security considerations:

Client to mediator connections SHOULD protect against MITM (pinning, etc)
Client to origin / redeemer could be intercepted to have the MITM harvest tokens

### LINKED ISSUE #44: Include Mediator's validation of keys in Security Considerations
We already have this text around Mediators validating keys; this should be referenced and expanded upon in Security Considerations.
```
Note that Mediators need to be careful in cases of key rotation; particularly,
an Origin could collude with an Issuer to try to rotate the key for each new Client in
order to link the client activity. The policy Mediators apply for key validation ought
to take such attacks into consideration.
```

tfpauly:
Closes #44 
Closes #19 
Closes #14

### DIFF
--- a/draft-private-access-tokens.md
+++ b/draft-private-access-tokens.md
@@ -498,10 +498,8 @@ matches a known public key for the Issuer. For example, the Mediator can fetch t
 key using the API defined in {{setup}}. This check is done to help ensure that the Client
 has not been given a unique key that could allow the Issuer to fingerprint or target
 the Client. If the key does not match, the Mediator rejects the request with an HTTP
-400 error. Note that Mediators need to be careful in cases of key rotation; particularly,
-an Origin could collude with an Issuer to try to rotate the key for each new Client in
-order to link the client activity. The policy Mediators apply for key validation ought
-to take such attacks into consideration.
+400 error. Note that Mediators need to be careful in cases of key rotation; see
+{{privacy-considerations}}.
 
 If the Mediator detects a version in the AccessTokenRequest that it does not recognize
 or support, it MUST reject the request with an HTTP 400 error.
@@ -762,8 +760,9 @@ Client activity could be linked if an Origin and Issuer collude to have unique k
 at specific Clients or sets of Clients. In order to mitigate this risk, the Mediator can 
 observe and validate the key_id presented by the Client to the Issuer. As described in
 {{issuance}}, Mediators MUST validate that the key_id in the Client's AccessTokenRequest
-matches a known public key for the Issuer. This validation allows for key rotation across
-all Clients, but not allow for per-Client keys.
+matches a known public key for the Issuer. The Mediator needs to support key rotation,
+but ought to disallow very rapid key changes, which could indicate that an Origin is colluding
+with an Issuer to try to rotate the key for each new Client in order to link the client activity.
 
 ## Issuer and Mediator Ownership
 
--- a/draft-private-access-tokens.md
+++ b/draft-private-access-tokens.md
@@ -735,12 +735,14 @@ All communication channels MUST use server-authenticated HTTPS. Where appropriat
 Clients and Mediators, connections MAY mutually authenticate both client and server, or use mechanisms
 such as TLS certificate pinning, to mitigate the risk of channel compromise.
 
-An attacker than can intermediate the channel between Client and Origin can
+An attacker that can intermediate the channel between Client and Origin can
 observe a TokenChallenge, and can view a Token being presented for authentication
-to an Origin. Origins can avoid this Token being used for some future connection
-by the attacker impersonating the Client by ensuring that the redemption_nonce
-presented in the TokenChallenge is bound to the specific TLS session with the
-Client.
+to an Origin. Scoping the TokenChallenge nonce to the Client HTTP session prevents
+Tokens being collected in one session and then presented to the Origin in another. 
+Note that an Origin cannot distinguish between a connection to a single Client and 
+a connection to an attacker intermediating multiple Clients. Thus, it is possible for
+an attacker to collect and later present Tokens from multiple clients over the same 
+Origin session.
 
 # Privacy Considerations {#privacy-considerations}
 
@@ -761,7 +763,7 @@ at specific Clients or sets of Clients. In order to mitigate this risk, the Medi
 observe and validate the key_id presented by the Client to the Issuer. As described in
 {{issuance}}, Mediators MUST validate that the key_id in the Client's AccessTokenRequest
 matches a known public key for the Issuer. This validation allows for key rotation across
-all Clients, but not allow for per-Client targetted keys.
+all Clients, but not allow for per-Client keys.
 
 ## Issuer and Mediator Ownership
 
--- a/draft-private-access-tokens.md
+++ b/draft-private-access-tokens.md
@@ -727,14 +727,15 @@ signal, such as an IP address, especially since malicious clients can generate g
 Private Access Tokens and for Origins to work. However, similar DoS vectors already exist
 for Origins, e.g., at the underlying TLS layer.
 
-## Man-in-the-Middle Attacks
+## Channel Security
 
-An attacker that can act as a man-in-the-middle for Mediator/Issuer communication can
+An attacker that can act as an intermediate between Mediator and Issuer communication can
 influence or disrupt the ability for the Issuer to correctly rate-limit token issuance.
-Communication between Mediators and Issuers MUST use a secure HTTPS connection, and SHOULD
-prevent man-in-the-middle attacks by employing an approach like TLS certificate pinning.
+All communication channels MUST use server-authenticated HTTPS. Where appropriate, e.g., between
+Clients and Mediators, connections MAY mutually authenticate both client and server, or use mechanisms
+such as TLS certificate pinning, to mitigate the risk of channel compromise.
 
-An attacker than can act as a man-in-the-middle for Client/Origin communication can
+An attacker than can intermediate the channel between Client and Origin can
 observe a TokenChallenge, and can view a Token being presented for authentication
 to an Origin. Origins can avoid this Token being used for some future connection
 by the attacker impersonating the Client by ensuring that the redemption_nonce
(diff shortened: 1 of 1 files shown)


draft-private-access-tokens.md
```diff
@@ -727,9 +727,41 @@ signal, such as an IP address, especially since malicious clients can generate g
 Private Access Tokens and for Origins to work. However, similar DoS vectors already exist
 for Origins, e.g., at the underlying TLS layer.
 
+## Man-in-the-Middle Attacks
```
chris-wood, line 730:
```suggestion
## Channel Security
```

draft-private-access-tokens.md
```diff
@@ -727,9 +727,41 @@ signal, such as an IP address, especially since malicious clients can generate g
...
+
+An attacker that can act as a man-in-the-middle for Mediator/Issuer communication can
```
chris-wood, line 732:
```suggestion
An attacker that can act as a intermediate Mediator or Issuer communication via impersonation or any other means can
```

draft-private-access-tokens.md
```diff
@@ -727,9 +727,41 @@ signal, such as an IP address, especially since malicious clients can generate g
...
+influence or disrupt the ability for the Issuer to correctly rate-limit token issuance.
+Communication between Mediators and Issuers MUST use a secure HTTPS connection, and SHOULD
+prevent man-in-the-middle attacks by employing an approach like TLS certificate pinning.
```
chris-wood, line 735:
```suggestion
All communication channels MUST use server-authenticated HTTPS. Where appropriate, e.g., between
Clients and Mediators, connections MAY mutually authenticate both client and server, or use mechanisms
such as TLS certificate pinning, to mitigate the risk of channel compromise.
```

draft-private-access-tokens.md
```diff
@@ -727,9 +727,41 @@ signal, such as an IP address, especially since malicious clients can generate g
...
+
+An attacker than can act as a man-in-the-middle for Client/Origin communication can
```
chris-wood, line 737:
```suggestion
An attacker than can intermediate the channel between Client and Origin can
```

draft-private-access-tokens.md
```diff
@@ -727,9 +727,41 @@ signal, such as an IP address, especially since malicious clients can generate g
...
+observe a TokenChallenge, and can view a Token being presented for authentication
+to an Origin. Origins can avoid this Token being used for some future connection
+by the attacker impersonating the Client by ensuring that the redemption_nonce
+presented in the TokenChallenge is bound to the specific TLS session with the
+Client.
```
chris-wood, line 742:
```suggestion
to an Origin. The protocol in this document does not allow Origins to protect against this type of attack.

[OPEN ISSUE: unless both client and server independently derive the challenge from the underlying channel, this problem will always remain.]
```

draft-private-access-tokens.md
```diff
@@ -727,9 +727,41 @@ signal, such as an IP address, especially since malicious clients can generate g
...
+matches the origin that is providing the HTTP authentication challenge, where the matching logic
+is defined for same-origin policies in {{?RFC6454}}. Clients MAY further limit which
+authentication challenges they are willing to respond to, for example by only accepting
+challenges when the origin is a web site to which the user navigated.
+
+## Client Identification with Unique Keys
+
+Client activity could be linked if an Origin and Issuer collude to have unique keys targeted
+at specific Clients or sets of Clients. In order to mitigate this risk, the Mediator is able
+to see and validate the key_id presented by the Client to the Issuer. As described in
```
chris-wood, line 760:
```suggestion
at specific Clients or sets of Clients. In order to mitigate this risk, the Mediator can 
observe and validate the key_id presented by the Client to the Issuer. As described in
```

draft-private-access-tokens.md
```diff
@@ -727,9 +727,41 @@ signal, such as an IP address, especially since malicious clients can generate g
...
+{{issuance}}, Mediators MUST validate that the key_id in the client's AccessTokenRequest
```
chris-wood, line 761:
```suggestion
{{issuance}}, Mediators MUST validate that the key_id in the Client's AccessTokenRequest
```

draft-private-access-tokens.md
```diff
@@ -727,9 +727,41 @@ signal, such as an IP address, especially since malicious clients can generate g
...
+matches a known public key for the Issuer. This validation should allow for key rotation across
```
chris-wood, line 762:
```suggestion
matches a known public key for the Issuer. This validation allows for key rotation across
```

chris-wood (changes requested):
Thanks! Looks good modulo a couple suggestions and one point about the challenge.

draft-private-access-tokens.md
tfpauly:
```suggestion
An attacker that can act as an intermediate between Mediator and Issuer communication can
```

chris-wood: 🤦

draft-private-access-tokens.md
tfpauly: @chris-wood hm, I'd like to have a bit of a better answer here rather than declaring we can't protect against it.

The main point, from the issue, was not to prevent the client-origin communication from having an intermediary watching things, but to stop the intermediary from harvesting tokens that it could then use in _other_ connections/sessions/contexts as its own client.

If the origin can require that the challenge/nonce is specific to a given session, can't it prove that this wasn't a token harvested on behalf of an unrelated connection?

chris-wood: > If the origin can require that the challenge/nonce is specific to a given session, can't it prove that this wasn't a token harvested on behalf of an unrelated connection?

When the connection is MITM'd, the session is with the attacker, not the client, so this just doesn't seem to work. A workaround would be to derive the nonce _from_ the connection, a la via [the Export API](https://datatracker.ietf.org/doc/html/rfc8446#section-7.5), but that's pretty uncommon. Things like WebAuthn have [slots for this type of binding](https://www.w3.org/TR/webauthn-2/#dom-collectedclientdata-tokenbinding) but it's basically never used in practice.

tfpauly: Right, the session is with the attacker, and then the client behind the attacker. Ah, so you’re saying the attacker could have other clients that it’s trying to squash together into one session?

janaiyengar: Do I understand correctly that this is a machine-on-the-side attacker, not a MITM attacker? AIUI, the attacker:
- watches the challenge and the token go by, harvesting both; and
- later spoofs a client to reach the origin and uses a token that matches the origin's challenge.

If my understanding is correct, then the crux of the issue is the risk that the nonce might repeat (which makes it a bad nonce.)

If so, it seems to me that @tfpauly's text suggests that an Origin can make a strong nonce by deriving it from something unique to the session as seen by the Origin (for example: timestamp + H(session key) + H(some client info)). I could of course be misunderstanding the issue and the text :-)

janaiyengar (commented):
A few comments

draft-private-access-tokens.md
```diff
@@ -727,9 +727,42 @@ signal, such as an IP address, especially since malicious clients can generate g
 Private Access Tokens and for Origins to work. However, similar DoS vectors already exist
 for Origins, e.g., at the underlying TLS layer.
 
+## Channel Security
+
+An attacker that can act as an intermediate between Mediator and Issuer communication can
```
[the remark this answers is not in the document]
chris-wood: Intermediation is what happens when the attacker is in the middle and is the thing we're concerned with here, so I'd be inclined to stick with what's here.

draft-private-access-tokens.md
```diff
@@ -727,9 +727,42 @@ signal, such as an IP address, especially since malicious clients can generate g
...
+influence or disrupt the ability for the Issuer to correctly rate-limit token issuance.
+All communication channels MUST use server-authenticated HTTPS. Where appropriate, e.g., between
+Clients and Mediators, connections MAY mutually authenticate both client and server, or use mechanisms
+such as TLS certificate pinning, to mitigate the risk of channel compromise.
```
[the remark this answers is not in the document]
chris-wood: Network attackers can also just block connections to mediators or issuers, so I'm not sure client auth helps for that particular problem.

draft-private-access-tokens.md
chris-wood: @tfpauly I updated my suggestion to try and make the limitation (and open problem) more clear.

chris-wood: @janaiyengar the attacker needs to intercept TLS to see the challenge and resulting Token, so it's not enough to be on the side, but maybe I misunderstood you?

draft-private-access-tokens.md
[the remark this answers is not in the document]
dvorak42: The concern might be if a party pretending to be the mediator asserts something about some client ID causing the issuer to think they've hit some policy limit. Though similarly a malicious mediator can do the same.

draft-private-access-tokens.md
dvorak42: Does the extra ability to present multiple tokens in the same session, versus the attacker just mitming the individual connections and intercepting/replaying the tokens separately there introduce much of an issue? Though you can sort of solve both with the exporter secret solution from the TLS session.

draft-private-access-tokens.md
```diff
@@ -727,9 +727,42 @@ signal, such as an IP address, especially since malicious clients can generate g
...
+authentication challenges they are willing to respond to, for example by only accepting
+challenges when the origin is a web site to which the user navigated.
+
+## Client Identification with Unique Keys
+
+Client activity could be linked if an Origin and Issuer collude to have unique keys targeted
+at specific Clients or sets of Clients. In order to mitigate this risk, the Mediator can 
+observe and validate the key_id presented by the Client to the Issuer. As described in
+{{issuance}}, Mediators MUST validate that the key_id in the Client's AccessTokenRequest
+matches a known public key for the Issuer. This validation allows for key rotation across
```
dvorak42, line 763: Probably as a separate issue this probably needs to be more nuanced. The issuer could decide to do a key rotation when a specific user/event happens (tracking the first 100 users vs the next set of user).

Mediators may almost need their own key rotation policy (or they delegate that to whatever known key service they use).

draft-private-access-tokens.md
```diff
@@ -727,9 +727,42 @@ signal, such as an IP address, especially since malicious clients can generate g
...
+all Clients, but not allow for per-Client targetted keys.
```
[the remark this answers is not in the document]
tfpauly: Moved the key rotation down to privacy considerations

draft-private-access-tokens.md
tfpauly: Modified the text a bit on Jana's suggestion, but yes, we can continue to improve it.

draft-private-access-tokens.md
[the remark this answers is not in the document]
tfpauly: Yeah, I don't know if we want specific client auth, beyond however the client would naturally authenticate to the mediator (device attestation, etc).

chris-wood (approved):

PR #47, opened 2021-09-24
Describe how to instantiate motivating use cases

chris-wood:
I don't love the geolocation exchange here. Mediators probably should not send anything that client's don't want them to send, so maybe that means clients opt-in to allowing their geohash to be forwarded. And maybe they decide to do that if Origins request tokens that need to be region specific? (Origins could maybe send that via the Accept-CH header in their challenge, and Clients could use that to send to the Mediator? I dunno.)

I also cut Client Linking as a use case.

### DIFF
--- a/draft-private-access-tokens.md
+++ b/draft-private-access-tokens.md
@@ -663,11 +663,11 @@ budget was exceeded.
 ## Client Geo-Location {#implement-geolocation}
 
 To instantiate this use case, the Issuer has an issuing key pair per geographic region, i.e.,
-each region has a unique policy key. The Mediator forwards information about the Client's
-geographic location in a "Sec-CH-Geohash" header to the Issuer during issuance. The Issuer
-uses this location information to select the correct policy key. The number of key pairs
-is then the cross product of the number of Origins that require per-region keys and the number
-of regions.
+each region has a unique policy key. When verifying the key for the Client request, the Mediator
+obtains the per-region key from the Issuer based on the Client's perceived location. During
+issuance, the Mediator checks that this key matches that of the Client's request. If it matches,
+the Mediator forwards the request to complete issuance. The number of key pairs is then the cross
+product of the number of Origins that require per-region keys and the number of regions.
 
 During redemption, Clients present their geographic location to Origins in a "Sec-CH-Geohash"
 header. Origins use this to obtain the appropriate policy verification key. Origins request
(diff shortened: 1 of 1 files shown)


draft-private-access-tokens.md
```diff
@@ -508,7 +494,8 @@ or support, it MUST reject the request with an HTTP 400 error.
 
 Before forwarding the Client's request to the Issuer, the Mediator adds headers
 listing both the ANON_CLIENT_ID as "Anonymous-Client-ID", and the ANON_ORIGIN_ID_PRIME as
-"Anonymous-Origin-ID".
+"Anonymous-Origin-ID". The Mediator MAY add additional context information about Clients,
+such as their geographic region encapsulated in a "Sec-CH-Geohash" header {{!I-D.ietf-geohash-hint}}.
```
tfpauly, line 498: I'd leave this out for now from this part and only mention it below.

```suggestion
"Anonymous-Origin-ID". The mediator MAY also add additional context information, but MUST
NOT add information that will uniquely identify a client.
```

draft-private-access-tokens.md
```diff
@@ -129,21 +129,7 @@ the available content it can serve based on the client's geographical region.
 Origins almost exclusively use client IP addresses for this purpose.
 
 
-## Client Linking
-
-An origin wishes to link multiple accesses from a client. The server does not
```
tfpauly, line 134: Can we mention this in privacy considerations, and say that a Client replaying tokens (and origins accepting that) would allow this happen again.

draft-private-access-tokens.md
```diff
@@ -692,20 +660,30 @@ the Client was able to request a token for the given (CLIENT_ID, ORIGIN_NAME) tu
 its budget. Failure to present a token can be interpreted as a signal that the client's token
 budget was exceeded.
 
-## Client Linking
+## Client Geo-Location {#implement-geolocation}
 
-In this example, a site wishes to determine if multiple connections from a user are from
-one user, or multiple. It is not important to map a connection to a non-anonymous user identifier.
+To instantiate this use case, the Issuer has an issuing key pair per geographic region, i.e.,
+each region has a unique policy key. The Mediator forwards information about the Client's
```
tfpauly, line 666: I don't know if the mediator would be sending the geohash to the Issuer here—instead, it should have a URI to look up on an issuer what key_id should be expected for each region, so it can verify the key is correct.

tfpauly (changes requested):

draft-private-access-tokens.md
chris-wood: Ah, that's much better!

draft-private-access-tokens.md
chris-wood: I'll address this in a followup PR.

tfpauly (approved):

PR #63, opened 2021-10-02
Private Access Token protocol rework

tfpauly:
Closes #55
Closes #58

This is a very large rework of the protocol, as I discussed with @chris-wood and @janaiyengar. We can discuss more in detail, and motivation, but this solves some major issues (like #55) as well as the ability for Origins and Issuers to use timing collusion to identify clients.

PR #68, opened 2021-10-08
Second pass at protocol rework

tfpauly:
Branched off of #63 

Removes groups, replaces it with new verification checks to ensure that clients cannot forge values


draft-private-access-tokens.md
```diff
@@ -408,23 +414,83 @@ Token matches the version in the TokenChallenge.
...
+
+- A counter of successful tokens issued
+- Whether or not a previous request was rejected by the Issuer  
+- The last received ANON_ISSUER_ORIGIN_ID value for this ANON_ORIGIN_ID, if any
+
+### Issuer State {#issuer-state}
+
+Issuers maintain a stable ISSUER_SECRET that they use in calculating values returned
+to the Mediator. If this value changes, it will open up a possibility for Clients to
+request extra tokens for an Origin without being limited, within a policy window.
```
dvorak42, line 481: Should this be rotating as part of the policy window?

draft-private-access-tokens.md
```diff
@@ -627,52 +727,28 @@ aad = concat(encode(1, keyID),
              encode(2, kemID),
              encode(2, kdfID),
              encode(2, aeadID),
-             encode(len(token_request), token_request))
+             encode(len(blinded_req), blinded_req),
```
dvorak42, line 730: Should this also pull in the version in the request? (to avoid collisions between future encrypted names (though only really an issue for long lived keys).

draft-private-access-tokens.md
```diff
@@ -568,22 +645,45 @@ blind_sig = rsabssa_blind_sign(skP, AccessTokenRequest.blinded_req)
...
+~~~
 
-The Issuer then updates its local state for the ANON_CLIENT_ID, ensuring that the
-ORIGIN_NAME and ANON_ORIGIN_ID_PRIME values are stored if not already present, and
-the history of token issuance is recorded with sufficient granularity to apply the
-Issuer's policy. For example, if the policy is simply meant to bound the number of
-tokens given to a specific Client within the policy window, the Issuer needs to increment
-its counter of the the number of tokens issued for the ANON_CLIENT_ID.
+If the "Origin-Tag" is missing, or if the same ANON_ISSUER_ORIGIN_ID was previously
+received in a response for a different ANON_ORIGIN_ID within the same policy window,
```
dvorak42, line 677: This search only needs to be against the ANON_ORIGIN_ID for this client, right? Otherwise needing to search against all seen ANON_ORIGIN_ID pairs might become expensive.

draft-private-access-tokens.md
```diff
@@ -477,31 +556,45 @@ accept = message/access-token-response
...
-If the Mediator detects a version in the AccessTokenRequest that it does not recognize
-or support, it MUST reject the request with an HTTP 400 error.
+The Mediator finally checks to ensure that the AccessTokenRequest.request_tag value
+matches its own calculation of request_tag = HashToGroup(CLIENT_ID) * request_nonce.
+If the request_tag does not match, the Mediator rejects the request with an HTTP
+400 error.
+
+TODO: Chris to update this
+
+If the Mediator accepts the request, it will look up the state stored for this Client.
```
dvorak42, line 584: AFAIU, its a little odd that the mediator needs to look up the Client based on a "client ID" it has for the client (device identity, IP, etc), which is different from the new "CLIENT_ID" that the client has provided?

dvorak42 (commented):
Generally looks good, simpler than the group approach.

draft-private-access-tokens.md
tfpauly: @chris-wood ?

draft-private-access-tokens.md
tfpauly: Yes, it needs its own notion of the client...

draft-private-access-tokens.md
```diff
@@ -471,16 +478,15 @@ For each tuple of (Client, ANON_ORIGIN_ID, policy window), the Mediator maintain
...
 - The last received ANON_ISSUER_ORIGIN_ID value for this ANON_ORIGIN_ID, if any
 
 ### Issuer State {#issuer-state}
 
 Issuers maintain a stable ISSUER_SECRET that they use in calculating values returned
 to the Mediator. If this value changes, it will open up a possibility for Clients to
 request extra tokens for an Origin without being limited, within a policy window.
-
-TODO: Chris to describe how ISSUER_SECRET should be calculated
+ISSUER_SECRET
```
tfpauly, line 489: @chris-wood was this intended to describe how ISSUER_SECRET is calculated?

chris-wood: YEp :) Will push

draft-private-access-tokens.md
chris-wood: We should probably just bind as much as we can to the ciphertext. I'll update this to include the mapping stuff.

Issue #67, opened 2021-10-07
Token-Count tracking


chris-wood:
Origins could potentially abuse the policy limit and token count to track specific clients across sessions, which is an anti-goal of the protocol.

PR #74, opened 2021-10-11
Add discussion of tracking using client token state.

chris-wood:
First cut at #67.

### DIFF
--- a/draft-private-access-tokens.md
+++ b/draft-private-access-tokens.md
@@ -220,8 +220,8 @@ redemption is limited to the fact that this is a client that passed a
 Mediator's checks and has not exceeded the per-origin limit defined by an
 Issuer. Generally, if a user is willing to use Private Access Tokens with
 a particular origin (or all origins), there is no need for per-challenge
-user interaction (this assumes that Clients can either silently authenticate
-to Mediators, or have already authenticated previously).
+user interaction. Note that the Issuance flow may separately involve user interaction if
+the Mediator needs to authenticate the Client.
 
 Since tokens are issued using a separate connection through a Mediator
 to an Issuer, the process of issuance can add user-perceivable latency.


draft-private-access-tokens.md
```diff
@@ -745,14 +745,28 @@ such as TLS certificate pinning, to mitigate the risk of channel compromise.
...
 a connection to an attacker intermediating multiple Clients. Thus, it is possible for
-an attacker to collect and later present Tokens from multiple clients over the same 
+an attacker to collect and later present Tokens from multiple clients over the same
 Origin session.
 
 # Privacy Considerations {#privacy-considerations}
 
+## Client Token State and Origin Tracking
+
+Origins SHOULD NOT challenge clients for access tokens more than once in a given context,
```
dvorak42, line 758: Might be worth specifying what counts as a context, specifically whether it's limited to the policy window.

draft-private-access-tokens.md
```diff
@@ -745,14 +745,28 @@ such as TLS certificate pinning, to mitigate the risk of channel compromise.
 An attacker that can intermediate the channel between Client and Origin can
 observe a TokenChallenge, and can view a Token being presented for authentication
 to an Origin. Scoping the TokenChallenge nonce to the Client HTTP session prevents
-Tokens being collected in one session and then presented to the Origin in another. 
-Note that an Origin cannot distinguish between a connection to a single Client and 
+Tokens being collected in one session and then presented to the Origin in another.
```
tfpauly, line 748:
```suggestion
Tokens from being collected in one session and then presented to the Origin in another.
```

draft-private-access-tokens.md
tfpauly: It's very specifically not a policy window—there can be many challenges in a window. 

I'd suggest maybe phrasing this all as: "Origins SHOULD only generate token challenges based on client action, such as when a user loads a website. Clients SHOULD ignore token challenges if an Origin tries to force the client to present tokens multiple times without any new client-initiated action."

PR #76, opened 2021-10-12
Token key ID and rotation

### LINKED ISSUE #57: Issuer key rotation
The different entities that rely on the two Issuer keys are not synchronized in their fetching of the keys, and there are a few failure modes as a result. We should think about these failure modes.

You could have the Issuer respond with a key when it is presented a key_id. But that might have implications for tracking.

chris-wood:
Closes #57.

### DIFF
--- a/draft-private-access-tokens.md
+++ b/draft-private-access-tokens.md
@@ -1034,7 +1034,7 @@ this 'single issuer/mediator' fashion reduces the privacy promises to those of P
 This may be desirable for a redemption flow that is limited to specific issuers and mediators,
 but should be avoided where hiding origins from the mediator is desirable.
 
-# Deployment Considerations {$deploy}
+# Deployment Considerations {#deploy}
 
 # Issuer Key Rollout
 


draft-private-access-tokens.md
```diff
@@ -522,8 +526,10 @@ policy window.
 Issuers are expected to have the private key that corresponds to ISSUER_KEY,
 which allows them to decrypt the ORIGIN_NAME values in requests.
 
-Issuers also need to know the current ORIGIN_TOKEN_KEY public key and corresponding
-private key, for each ORIGIN_NAME that is served by the Issuer.
+Issuers also need to know the set of valid ORIGIN_TOKEN_KEY public keys and corresponding
+private key, for each ORIGIN_NAME that is served by the Issuer. Origins SHOULD update
+their view of the ORIGIN_TOKEN_KEY regularly to ensure that Client requests do not fail
+after ORIGIN_TOKEN_KEY rotation.
```
smhendrickson, line 532: Should we be more strict here?

Issuers should publish their expected rotation schedule, which would include the hard deadline origins must pick up new keys by.

draft-private-access-tokens.md
```diff
@@ -616,16 +623,20 @@ The structure fields are defined as follows:
...
 This document defines version 1.
 
-- "key_id" is a collision-resistant hash that identifies the ISSUER_KEY public
-key, generated as SHA256(KeyConfig).
-
 - "mapping_generator", "mapping_key", and "mapping_proof" are computed as described above.
 
-- "encrypted_origin_name" is an encrypted structure that contains ORIGIN_NAME,
-calculated as described in {{encrypt-origin}}.
+- "token_key_id" is the least significant byte of the ORIGIN_TOKEN_KEY key ID, which is
```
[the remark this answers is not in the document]
chris-wood: Sure, 8 was just a placeholder. We can probably get away with one or two.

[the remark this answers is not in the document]
chris-wood: I'd recommend keeping it at 8 bits for now and then letting the WG bash out the number of bits we care about.

draft-private-access-tokens.md
```diff
@@ -451,6 +449,12 @@ key from the Issuer, and validate that the message matches the hash of the origi
 TokenChallenge for this session, SHA256(TokenChallenge), and that the version of the
 Token matches the version in the TokenChallenge.
 
+If a Client's issuance request fails with a 401 error, as described in {{request-two}},
+the Client MUST react to the challenge as if it could not produce a valid Authorization
+response.
```
janaiyengar, line 454: Normally, this would be a server workflow, would it not? Clients may not have a workflow here... perhaps the Client could respond with an empty response and allow the Server to follow the auth failure workflow as necessary?

draft-private-access-tokens.md
```diff
@@ -596,18 +602,19 @@ mapping_generator = SerializeElement(blind_generator)
 mapping_proof = SerializeProof(key_proof)
 ~~~
 
-The Client then constructs a Private Access Token request using blinded_req,
-mapping_key, mapping_generator, mapping_proof, and request_tag:
+The Client then constructs a Private Access Token request using mapping_key,
+mapping_generator, mapping_proof, blinded_req, and the origin.
```
janaiyengar, line 606:
```suggestion
mapping_generator, mapping_proof, blinded_req, and origin information.
```

draft-private-access-tokens.md
```diff
@@ -693,16 +704,17 @@ sec-token-count = 3
...
 - The "Sec-Token-Count" header is present
 - The AccessTokenRequest contains a supported version
-- For version 1, the AccessTokenRequest.key_id corresponds to the ID of the ISSUER_KEY held by the Issuer
+- For version 1, the AccessTokenRequest.name_key_id corresponds to the ID of the ISSUER_KEY held by the Issuer
 - For version 1, the AccessTokenRequest.encrypted_origin_name can be decrypted using the
 Issuer's private key (the private key associated with ISSUER_KEY), and matches
 an ORIGIN_NAME that is served by the Issuer
 - For version 1, the AccessTokenRequest.blinded_req is of the correct size
+- For version 1, the AccessTokenRequest.token_key_id corresponds to an ID of an ORIGIN_TOKEN_KEY
+for the corresponding ORIGIN_NAME
```
janaiyengar, line 717:
```suggestion
- For version 1, the AccessTokenRequest.token_key_id corresponds to the ID of the ORIGIN_TOKEN_KEY
for the corresponding ORIGIN_NAME
```
... unless I'm missing something

draft-private-access-tokens.md
```diff
@@ -1020,6 +1034,12 @@ this 'single issuer/mediator' fashion reduces the privacy promises to those of P
 This may be desirable for a redemption flow that is limited to specific issuers and mediators,
 but should be avoided where hiding origins from the mediator is desirable.
 
+# Deployment Considerations {$deploy}
```
janaiyengar, line 1037:
```suggestion
# Deployment Considerations {#deploy}
```

draft-private-access-tokens.md
chris-wood: Yep, that's fine too!

PR #85, opened 2021-10-15
Clarify requirements around endpoint auth

### LINKED ISSUE #48: Do we need mutual auth between Mediator and Issuer?
An issuer might need to ensure that a mediator is authenticated / strongly identified, since the anonymous client ID could be spoofed by another mediator. If someone pretending to be a mediator knew the ANON_CLIENT_ID and the ANON_ORIGIN_ID_PRIME, they could act as the mediator. This might be enough, but most mediator/issuer pairs likely would do mutual auth anyway.

janaiyengar:
Clarifies and proposes requirements on endpoint authentication. Also moves the requirements out of Security Considerations to up in the body of the draft.

Closes #48.

### DIFF
--- a/draft-private-access-tokens.md
+++ b/draft-private-access-tokens.md
@@ -785,8 +785,8 @@ forwarding it to the Issuer.
 ### Mediator-to-Issuer Request {#request-two}
 
 The Mediator and the Issuer MUST use a secure and Issuer-authenticated HTTPS
-connection. Since an Issuer issues tokens to only known Mediators, connections
-between them MUST use mutual authentication. They MAY additionally use
+connection. Also, Issuers MUST authenticate Mediators, either via mutual
+TLS or another form of application-layer authentication. They MAY additionally use
 mechanisms such as TLS certificate pinning, to mitigate the risk of channel
 compromise; see {{sec-considerations}} for additional about this channel.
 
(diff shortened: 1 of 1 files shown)


draft-private-access-tokens.md
```diff
@@ -783,6 +784,12 @@ forwarding it to the Issuer.
 
 ### Mediator-to-Issuer Request {#request-two}
 
+The Mediator and the Issuer MUST use a secure and Issuer-authenticated HTTPS
+connection. Since an Issuer issues tokens to only known Mediators, connections
+between them MUST use mutual authentication. They MAY additionally use
```
chris-wood, line 789:
```suggestion
The Mediator and the Issuer MUST use a secure and Issuer-authenticated HTTPS
connection. Also, Issuers MUST authenticate Mediators, either via mutual
TLS or another form of application-layer authentication. They MAY additionally use
```

draft-private-access-tokens.md
```diff
@@ -1070,11 +1077,13 @@ for Origins, e.g., at the underlying TLS layer.
...
-All communication channels MUST use server-authenticated HTTPS. Where appropriate, e.g., between
-Clients and Mediators, connections MAY mutually authenticate both client and server, or use mechanisms
-such as TLS certificate pinning, to mitigate the risk of channel compromise.
+An attacker that can act as an intermediate between Mediator and Issuer
+communication can influence or disrupt the ability for the Issuer to correctly
+rate-limit token issuance.  All communication channels use server-authenticated
+HTTPS. Where appropriate, e.g., between a Mediator and an Issuer, connections
+might require or choose mutual authentication between both endpoints or use
+mechanisms such as TLS certificate pinning, to mitigate the risk of channel
+compromise.
```
chris-wood, line 1086:
```suggestion
HTTPS. Some connections, e.g., between a Mediator and an Issuer, require
mutual authentication between both endpoints. Where appropriate, endpoints 
MAY use further enhancements such as TLS certificate pinning to mitigate 
the risk of channel compromise.
```

PR #89, opened 2021-10-19
Client state loss

smhendrickson:
Describe client state loss procedures

Tried to centralize this within the state requirements, and mentioned in the client-to-mediator flow.

### DIFF
--- a/draft-private-access-tokens.md
+++ b/draft-private-access-tokens.md
@@ -509,8 +509,8 @@ to identify a Client.
 Mediators must enforce that clients don't change their CLIENT_ID frequently, to ensure clients can't
 regularily evade the per-client policy as seen by the issuer. Mediators MUST not allow clients to
 change their CLIENT_ID more than once within a policy window, or in the subsequent policy window
-after a previous CLIENT_ID change. Alternatively, schemes where the mediator stores the encrypted
-(CLIENT_ID, CLIENT_SECRET) tuple on behalf of the client are possble, but not outlined here.
+after a previous CLIENT_ID change. Alternative schemes where the mediator stores the encrypted
+(CLIENT_ID, CLIENT_SECRET) tuple on behalf of the client are possble but not described here.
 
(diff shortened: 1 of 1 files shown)


draft-private-access-tokens.md
```diff
@@ -642,13 +646,19 @@ of identifying a Client is specific to each Mediator, and is not defined in this
 As examples, the Mediator could use device-specific certificates or account authentication
 to identify a Client.
 
+Mediators must enforce that clients don't change their CLIENT_ID frequently, to ensure clients can't
+regularily evade the per-client policy as seen by the issuer. Mediators MUST not allow clients to
+change their CLIENT_ID more than once within a policy window, or in the subsequent policy window
+after a previous CLIENT_ID change. Alternative schemes where the mediator stores the encrypted
+(CLIENT_ID, CLIENT_SECRET) tuple on behalf of the client are possble but not described here.
```
tfpauly, line 653:
```suggestion
Mediators must enforce that Clients don't change their CLIENT_ID frequently, to ensure Clients can't
regularily evade the per-client policy as seen by the issuer. Mediators MUST NOT allow Clients to
change their CLIENT_ID more than once within a policy window, or in the subsequent policy window
after a previous CLIENT_ID change. Alternative schemes where the mediator stores the encrypted
(CLIENT_ID, CLIENT_SECRET) tuple on behalf of the client are possble but not described here.
```

draft-private-access-tokens.md
```diff
@@ -828,6 +838,10 @@ If the Mediator has stored state that a previous request for this ANON_ORIGIN_ID
 rejected by the Issuer in the current policy window, it SHOULD reject the request without
 forwarding it to the Issuer.
 
+If the Mediator detects this client has changed their CLIENT_ID more frequently than allowed
```
tfpauly, line 841:
```suggestion
If the Mediator detects this Client has changed their CLIENT_ID more frequently than allowed
```

tfpauly:
Looks good, thanks @smhendrickson !

PR #91, opened 2021-10-21
note mediator/origin timestamp collusion

smhendrickson:
Note mediator/origin collusion.

This makes me wonder if allowing both interactive and non-interactive retains utility. While there are many deployments that can use this as is, avoiding timestamp correlation may be desirable for others. Enough so to warrent 'offline' token issuance and caching.

Thoughts?

### DIFF
--- a/draft-private-access-tokens.md
+++ b/draft-private-access-tokens.md
@@ -1215,7 +1215,7 @@ this 'single issuer/mediator' fashion reduces the privacy promises to those of t
 desirable for a redemption flow that is limited to specific issuers and mediators,
 but should be avoided where hiding origins from the mediator is desirable.
 
-If a Mediator and Origin collude, they could correlate a client's identity and origin access patterns through timestamp correlation. The
+If a Mediator and Origin are able to collude, they can correlate a client's identity and origin access patterns through timestamp correlation. The
 timing of a request to an Origin and subsequent token issuance to a Mediator can reveal the Client
 identity (as known to the Mediator) to the Origin, especially if repeated over multiple accesses.
 


draft-private-access-tokens.md
```diff
@@ -1211,6 +1211,10 @@ this 'single issuer/mediator' fashion reduces the privacy promises to those of t
 desirable for a redemption flow that is limited to specific issuers and mediators,
 but should be avoided where hiding origins from the mediator is desirable.
 
+A mediator and origin can collude to reveal the client identity through timestamp correlation. The
+timing of a request to an origin and subsiquent token issuance to a mediator can reveal the client
```
tfpauly, line 1215:
```suggestion
timing of a request to an Origin and subsequent token issuance to a Mediator can reveal the Client
```

draft-private-access-tokens.md
```diff
@@ -1211,6 +1211,10 @@ this 'single issuer/mediator' fashion reduces the privacy promises to those of t
...
+identity (as known to the mediator) to the origin, especially if repeated over multiple accesses.
```
tfpauly, line 1216:
```suggestion
identity (as known to the Mediator) to the Origin, especially if repeated over multiple accesses.
```

draft-private-access-tokens.md
```diff
@@ -1211,6 +1211,10 @@ this 'single issuer/mediator' fashion reduces the privacy promises to those of t
 desirable for a redemption flow that is limited to specific issuers and mediators,
 but should be avoided where hiding origins from the mediator is desirable.
 
+A mediator and origin can collude to reveal the client identity through timestamp correlation. The
```
tfpauly, line 1214:
```suggestion
A Mediator and Origin can collude to reveal a client's identity through timestamp correlation. The
```

draft-private-access-tokens.md
```diff
@@ -1211,6 +1211,10 @@ this 'single issuer/mediator' fashion reduces the privacy promises to those of t
 desirable for a redemption flow that is limited to specific issuers and mediators,
 but should be avoided where hiding origins from the mediator is desirable.
 
+A Mediator and Origin can collude to reveal a client's identity through timestamp correlation. The
```
tfpauly, line 1214:
```suggestion
If a Mediator and Origin collude, they could correlate a client's identity and origin access patterns through timestamp correlation. The
```

draft-private-access-tokens.md
```diff
@@ -1201,7 +1201,7 @@ a single valid public key for signing tokens at a time. The Client does not pres
 of the token public key to the Issuer, but instead expects the Issuer to infer the correct key based
 on the information the Issuer knows, specifically the origin_name itself.
 
-## Issuer and Mediator Ownership
+## Issuer, Mediator and Origin Ownership
```
tfpauly, line 1204: I'm not sure this makes sense to have "origin" under "ownership" here, since we talk only about collusion and not ownership for this case.

draft-private-access-tokens.md
```diff
@@ -1211,6 +1211,10 @@ this 'single issuer/mediator' fashion reduces the privacy promises to those of t
 desirable for a redemption flow that is limited to specific issuers and mediators,
 but should be avoided where hiding origins from the mediator is desirable.
 
+A Mediator and Origin can collude to reveal a client's identity through timestamp correlation. The
+timing of a request to an Origin and subsequent token issuance to a Mediator can reveal the Client
```
tfpauly, line 1215: I think we should focus on correlating identity and access... it doesn't directly reveal just the client to the origin, it allows the origin and mediator to compare notes.

tfpauly:
Another way to look at this is that we expect (Origin, Issuer) to have collaboration (if not necessarily shared ownership) and (Client, Mediator) to have collaboration. If those two halves merge, such that (Origin, Issuer, Mediator) are all sharing data, what's known to one is by definition known to all. So, if you tell your identity to a Mediator who also runs the Origin and Issuer you're accessing, you shouldn't expect them to not be able to share information.

Essentially, it degenerates to logging into the service directly. So the client may as well just authenticate to the origin, or treat it as if they did.

PR #108, opened 2021-12-06
Add type to WWW-Authenticate header.

### LINKED ISSUE #107: Tie the WWW-Authenticate around the keys and max-age.
Currently the issuer-key, token-key, and max-age are not versioned inside the WWW-Authenticate, meaning that either we need current parsing of a scheme to ignore unknown fields in the scheme challenge (the outermost layer defined in RFC7235), we have to introspect the challenge field to then figure out how to parse the outer keys/max-age fields, or we have to use a different scheme name whenever we want to change those.

One potential fix is to encapsulate the whole blob of data in an "outer" challenge structure:

~~~
struct {
    uint8_t version;
    TokenChallenge token_challenge;
    opaque issuer_key<1..2^16-1>;
    opaque token_key<1..2^16-1>;
    uint32_t max_age;
} PATAuthenticateChallenge;

WWW-Authenticate: PrivateAccessToken challenge=abc..., PrivateAccessToken challenge=def...
~~~

Where abc/def are PATAuthenticateChallenge(s).

Additionally we can potentially rename TokenChallenge to get rid of the extra layering.

dvorak42:
This adds a type to indicate the use of Private Access Tokens, and generalizes the scheme to "AnonymousCredential".

Other schemes, like PrivacyPass will share the scheme and use their own type.

Fixes #107.

### DIFF
--- a/draft-private-access-tokens.md
+++ b/draft-private-access-tokens.md
@@ -525,8 +525,13 @@ The structure fields are defined as follows:
 
 - "redemption_nonce" is a fresh 32-byte nonce generated for each redemption request.
 
-When used in an authentication challenge, the "PrivateAccessToken" scheme uses the
+When used in an authentication challenge, the "AnonymousCredential" scheme uses the
 following attributes:
+
+- "type", which contains the type of anonymous credential being used. For Private
+Acces Tokens, this is the string "PrivateAccessToken".
+
+The "PrivateAccessToken" type additionally has the following attributes:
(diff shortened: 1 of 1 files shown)


chris-wood (commented):
Nice 👍 :)

tfpauly:
Nice! I'm sure we can bike shed a bit, with the rest of the Privacy Pass WG as well...

PR #110, opened 2021-12-08
Generalize signatures and add issuance protocol details and some security considerations

chris-wood:
We can probably use some more details about replay attacks, but this should get us started.


draft-pauly-privacypass-auth-scheme.md
```diff
@@ -304,4 +380,24 @@ Pointer to specification text: {{challenge-redemption}} of this document
...
+* Nk: The length in bytes of an output authenticator
+* Reference: Where this algorithm is defined
+
+The initial contents for this registry are defined in the table below.
+
+| Value  | Name                   | Public | Nk  | Reference    |
+|:-------|:-----------------------|:-------|:----|:-------------|
+| 0x0000 | (reserved)             | N/A    | N/A | N/A          |
+| 0x0001 | Blind RSA, 2048        | Y      | 256 | TODO         |
+| 0x0002 | Blind RSA, 4096        | Y      | 512 | TODO         |
```
tfpauly, line 401: Do we need both?

draft-pauly-privacypass-auth-scheme.md
```diff
@@ -304,4 +380,24 @@ Pointer to specification text: {{challenge-redemption}} of this document
...
+| 0x0003 | VOPRF(P-384, SHA-384)  | N      | 256 | TODO         |
```
tfpauly, line 402: So PATs would be another entry in this table?

I wonder if the name should be a bit more descriptive than the signature type. Or I guess PATs would be something like "Rate-Limited Blind RSA"?

tfpauly (approved):

draft-pauly-privacypass-auth-scheme.md
chris-wood: Yeah, these are sort of placeholder names until we have the specific drafts to point to. "Blind RSA" isn't very helpful, for example.

dvorak42: Wonder if we should leave some space so that we can at least keep similar entries close together numerically?

Though I guess that devolves into a type/version separation again.

tfpauly: Yeah, numbers should be cheap... but we can add in batches.

We can fill this out as we write the three issuance protocol specs

chris-wood: Yeah -- let's just fill this stuff out later.

PR #113, opened 2021-12-14
Add public/private metadata flags to IANA registry

tfpauly:


chris-wood (approved):

smhendrickson:
LGTM

At least with RSA Blind Signatures, I think it is _always_ possible for the token to contain private metadata because what is signed is garbage from the signers perspective.

PR #116, opened 2021-12-16
Mostly editorial pass

janaiyengar:
I've also made some changes to normative text. Specifically,
- removed a MUST NOT on UI interaction, since that seems quite unnecessary and entirely untestable
- replaced a SHOULD NOT with a MUST NOT on dealing with too many challenges from the origin, and tightened up the language a bit. I could not see a good reason to do anything but a MUST here, and being more precise about the requirements helps.


chris-wood (approved):

draft-pauly-privacypass-auth-scheme.md
```diff
@@ -329,17 +331,18 @@ protocol may separately involve user interaction if the client needs to be newly
...
-token authentication more than once for a single website load, it can indicate
-that the server is not functioning correctly, or is trying to attack or overload
-the client or issuance servers. In such cases, the client SHOULD ignore
-redundant token challengers, or else alert the user.
+An origin MUST NOT issue more than one challenge per client request. If an
+origin issues a large number of challenges, such as more than once for each
+request, this can indicate that the origin is either not functioning correctly
+or is trying to attack or overload the client or issuance server. In such cases,
+a client MUST ignore redundant token challenges for the same request and SHOULD
+alert the user if possible.
```
tfpauly, line 345:
```suggestion
An origin MUST NOT issue more than one interactive challenge for a given token type
and issuer per client request. If an origin issues a large number of challenges, such
as more than once for each request, this can indicate that the origin is either not
functioning correctly or is trying to attack or overload the client or issuance server.
In such cases, a client MUST ignore redundant token challenges for the same
request and SHOULD alert the user if possible.

Origins MAY include multiple challenges, where each challenge refers to a different issuer
or a different token type, to allow clients to choose a preferred issuer or type.
```

tfpauly (approved):

PR #129, opened 2022-01-03
Try out moving the limit enforcement to the Attester

### LINKED ISSUE #125: Send policy from Issuer to Attester for enforcement
Currently, the Attester sends per-client state to the Issuer for the purposes of determining whether or not a given token request is permitted (or not). I see a couple problems with this:

1. The policy limit isn't necessarily secret from the Attester anyway, since it (a) knows the per-client counts and (b) can see when the Issuer refuses to produce a token response. It can then infer what the limit is. It could even act as a malicious client to learn the per-origin limit if it wanted.
2. The Issuer currently reveals when it refuses to process a token. We could change that by having it just randomize the token response in the event of failure, but then the Attester could just act as a client and learn whatever real clients learn, so this doesn't make much sense to me.
3. Bucketizing the counts from Attester to Issuer seems leaky at best, especially since the right size of the buckets seems to depend on client activity. I don't know of a reasonable strategy here that would allow us to state concrete properties about what the Issuer can and cannot learn per client. And it seems like we want something like k-anonymity?

I think we should instead have the Issuer send the Attester the per-origin policy limit, and then require the Attester to enforce the policy based on that limit. We already require origins to not have unique policies for privacy reasons, so this seems fine. Moreover, giving the Attester this policy information doesn't seem like revealing anything new. But it would mean that the origins now trust Attesters directly to enforce limits, rather than Issuers being the ones to do that. Since Attesters are already transitively trusted through Issuers, this doesn't seem like a stretch.

tfpauly:
Closes #125

### DIFF
--- a/draft-privacypass-rate-limit-tokens.md
+++ b/draft-privacypass-rate-limit-tokens.md
@@ -424,13 +424,14 @@ Its ABNF is:
     Sec-Token-Request-Blind = sf-binary
 ~~~
 
-The "Sec-Token-Count" is an Item Structured Header {{!RFC8941}}. Its
-value MUST be an Integer. This header is sent on Attester-to-Issuer
-requests ({{request-two}}), and contains the number of times a
-Client has previously received a token for an Origin. Its ABNF is:
+The "Sec-Token-Limit" is an Item Structured Header {{!RFC8941}}. Its
+value MUST be an Integer. This header is sent on Issuer-to-Attester
+responses ({{response-one}}), and contains the number of times a
+Client can retrieve a token for the requested Origin, set by the
+Issuer. Its ABNF is:
 
 ~~~
-    Sec-Token-Count = sf-integer
+    Sec-Token-Limit = sf-integer
 ~~~
 
 ## Client-to-Attester Request {#request-one}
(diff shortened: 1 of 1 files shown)


chris-wood (approved):
This looks great! I like the new 429 signal, too, as that allows clients to differentiate between "bad request" to "out of tokens."

draft-privacypass-rate-limit-tokens.md
```diff
@@ -424,13 +424,14 @@ Its ABNF is:
...
 ~~~
 
-The "Sec-Token-Count" is an Item Structured Header {{!RFC8941}}. Its
-value MUST be an Integer. This header is sent on Attester-to-Issuer
-requests ({{request-two}}), and contains the number of times a
-Client has previously received a token for an Origin. Its ABNF is:
+The "Sec-Token-Limit" is an Item Structured Header {{!RFC8941}}. Its
+value MUST be an Integer. This header is sent on Issuer-to-Attester
+responses ({{response-one}}), and contains the number of times a
+Client can retrieve a token for the requested Origin, set by the
```
smhendrickson, line 430: can retrieve a token for the requested Origin _within a policy window_, as set ..

smhendrickson (approved):
LGTM

tfpauly:
@chris-wood the 429 signal was there before, just moved in location.

draft-privacypass-rate-limit-tokens.md
```diff
@@ -424,13 +424,14 @@ Its ABNF is:
...
+Issuer. Its ABNF is:
```
tfpauly, line 431:
```suggestion
Client can retrieve a token for the requested Origin within a policy window,
as set by the Issuer. Its ABNF is:
```

PR #132, opened 2022-01-04
Fix and unify issue-key name across docs

### LINKED ISSUE #131: Rename "issuer-key"
The issuer's config value "issuer-key" is the base64-url encoded public key used for fetching tokens, but the name is a bit misleading. Maybe we can rename this to "token-issuance-key" or similar?

chris-wood:
Closes #131

### DIFF
--- a/draft-pauly-privacypass-auth-scheme.md
+++ b/draft-pauly-privacypass-auth-scheme.md
@@ -203,15 +203,15 @@ following attributes:
 value. This MUST be unique for every 401 HTTP response to prevent replay attacks.
 This attribute is required for all challenges.
 
-- "issuer-key", which contains a base64url encoding of the SubjectPublicKeyInfo object
-for use with the issuance protocol indicated by the challenge. This attribute MAY
+- "token-issuance-key", which contains a base64url encoding of the public key for
+use with the issuance protocol indicated by the challenge. This attribute MAY
 be omitted in deployments where clients are able to retrieve the issuer key using
 an out-of-band mechanism.
 
 - "max-age", an optional attribute that consists of the number of seconds for which
 the challenge will be accepted by the Origin.
 
-Clients can ignore the challenge if the issuer-key is invalid or otherwise untrusted.
+Clients can ignore the challenge if the token-issuance-key is invalid or otherwise untrusted.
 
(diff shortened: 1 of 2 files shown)


tfpauly:
Hm, I'm not sure I prefer "token-issuance-key" to "issuer-key". What's wrong with the issuer-key naming, since it's shorter and seems pretty clear still?

chris-wood:
My claim is that it's not really clear. The issuer has many keys in its possession, including origin name encryption key, token _issuance_ keys (private keys), token _validation_ keys (public keys), and so on. I think we ought to be specific with what this thing is. Maybe that's not "token-issuance-key" (since it's really the validation key), but it's certainly not "issuer-key".

tfpauly:
Commented on the issue. I still prefer the simpler name here. I agree that "token-issuance-key" isn't correct, since it sounds more like a private key. The key we're talking about is the common one that all clients and origins see for all token types, which is why I think a generic name like "issuer key" is appropriate.

draft-pauly-privacypass-auth-scheme.md
```diff
@@ -203,23 +203,23 @@ following attributes:
 value. This MUST be unique for every 401 HTTP response to prevent replay attacks.
 This attribute is required for all challenges.
 
-- "issuer-key", which contains a base64url encoding of the SubjectPublicKeyInfo object
-for use with the issuance protocol indicated by the challenge. This attribute MAY
+- "token-issuance-key", which contains a base64url encoding of the public key for
```
tfpauly, line 206: How about:

```suggestion
- "token-validation-key", which contains a base64url encoding of the public key for
```

Or

```suggestion
- "token-key", which contains a base64url encoding of the public key for
```

I may lean towards the short "token-key" since this is sent on the wire, and in the challenge, it's only the verification key that matters.

tfpauly (changes requested):

draft-pauly-privacypass-auth-scheme.md
chris-wood: Yeah, based on the other issue chat, I think token-validation-key is best. I'll update this soon to take that into account.

tfpauly:
Discussed on call, change to "token-key"

chris-wood:
Closing in favor of #138.

PR #139, opened 2022-01-11
Drop OHTTP dependency and redefine the name configuration

### LINKED ISSUE #112: Origin name encryption algorithm
The issuer's HPKE configuration admits multiple algorithm configurations, each one indicating a KEM, KDF, and AEAD tuple. Clients use this configuration when encrypting the origin name to the issuer. However, the client's request currently doesn't convey what algorithm selection was used for this encryption. We can either restrict things so that each HPKE configuration has exactly one algorithm tuple, or the request can convey the client's selection. (I don't think trial decryption is a reasonable option here.)

### LINKED ISSUE #124: Consider removing OHTTP dependency
We currently borrow the public configuration structure from OHTTP, and the just use it for HPKE operations. This dependency doesn't buy us much at this point as we're not using OHTTP, so we might as well inline the structure definition.

chris-wood:
The NameKey fully defines the algorithms and key used for encrypting origins, as well as the corresponding key ID. Clients use these keys entirely or not, meaning there's no client-side negotiation. Since the key ID fully defines the key and algorithms, clients do not need to convey anything beyond the key ID to issuers in their TokenRequest.

Closes #124.
Closes #112.

### DIFF
--- a/draft-privacypass-rate-limit-tokens.md
+++ b/draft-privacypass-rate-limit-tokens.md
@@ -243,11 +243,26 @@ Issuers MUST provide three parameters for configuration:
 1. Issuer Request URI: a token request URL for generating access tokens.
    For example, an Issuer URL might be https://issuer.example.net/token-request. This parameter
    uses resource media type "text/plain".
-1. Origin Name Key: a `KeyConfig` as defined in {{!OHTTP=I-D.thomson-http-oblivious}} to use when
-   encrypting the Origin Name in issuance requests. This parameter uses resource media type
-   "application/ohttp-keys".
+1. Origin Name Key: a `NameKey` structure as defined below to use when encrypting the Origin
+   Name in issuance requests. This parameter uses resource media type "application/pat-name-key".
+   The Npk parameter corresponding to the HpkeKdfId can be found in {{HPKE}}.
 
-These parameters can be obtained from an Issuer via a directory object, which is a JSON
+~~~
+opaque HpkePublicKey[Npk]; // defined in I-D.irtf-cfrg-hpke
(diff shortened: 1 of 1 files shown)


draft-privacypass-rate-limit-tokens.md
```diff
@@ -243,11 +243,26 @@ Issuers MUST provide three parameters for configuration:
 1. Issuer Request URI: a token request URL for generating access tokens.
    For example, an Issuer URL might be https://issuer.example.net/token-request. This parameter
    uses resource media type "text/plain".
-1. Origin Name Key: a `KeyConfig` as defined in {{!OHTTP=I-D.thomson-http-oblivious}} to use when
-   encrypting the Origin Name in issuance requests. This parameter uses resource media type
-   "application/ohttp-keys".
+1. Origin Name Key: a `NameKey` structure as defined below to use when encrypting the Origin
+   Name in issuance requests. This parameter uses resource media type "application/pat-name-key".
```
tfpauly, line 247: Can we not use "PAT" here, since it's not called private access tokens in this doc?

chris-wood: Oh, sure, I didn't feel strongly about the name beyond that it should relate to privacy pass in some way.

smhendrickson (approved):

draft-privacypass-rate-limit-tokens.md
tfpauly: That's fine with me, yeah!

chris-wood: Cool -- fixed!

tfpauly (approved):

PR #152, opened 2022-02-01
Use ECDSA instead of Ed25519

chris-wood:
cc @FredericJacobs


tfpauly (approved):

smhendrickson (approved):

FredericJacobs (approved):
Approving this direction, but need more security analysis before we submit this draft as a revision to DataTracker. (See #153)

PR #176, opened 2022-03-02
Explain extensions that could add metadata

### LINKED ISSUE #172: Rate-Limited Tokens: additional context information
Section 5.4 says 

> The Attester MAY also add additional context information, but MUST NOT add information that will uniquely identify a Client.

This sentence does not seem to prevent an attester to reduce the anonymity pool to (for example) 2 users, as they would not “add information that will uniquely identity a Client”. 

* What kind of additional context would be allowed? Do we need such additional context?
* If yes, is there a way to be more prescriptive in the wording to ensure that the above “attack” on the size of the anonymity is not doable?

tfpauly:
Closes #172

### DIFF
--- a/draft-privacypass-rate-limit-tokens.md
+++ b/draft-privacypass-rate-limit-tokens.md
@@ -614,9 +614,8 @@ If this fails, the Attester MUST return an HTTP 400 error to the Client.
 The Attester MUST NOT add information that will uniquely identify a Client,
 or associate the request with a small set of possible Clients. Extensions
 to this protocol MAY allow Attesters to add information that can be used
-to separate large populations (numbering in the thousands or more), such
-as providing information about what country or region to which a Client
-belongs.
+to separate large populations, such as providing information about the country
+or region to which a Client belongs.
 
 ~~~
 :method = POST


chris-wood (approved):

draft-privacypass-rate-limit-tokens.md
```diff
@@ -611,8 +611,12 @@ compromise; see {{sec-considerations}} for additional about this channel.
 Before copying and forwarding the Client's TokenRequest request to the Issuer,
 the Attester validates the Client's stable mapping request as described in {{attester-anon-issuer-origin-id}}.
 If this fails, the Attester MUST return an HTTP 400 error to the Client.
-The Attester MAY also add additional context information, but MUST NOT add information
-that will uniquely identify a Client.
+The Attester MUST NOT add information that will uniquely identify a Client,
+or associate the request with a small set of possible Clients. Extensions
+to this protocol MAY allow Attesters to add information that can be used
+to separate large populations (numbering in the thousands or more), such
```
dvorak42, line 617: Maybe leave the size of a "large population" unspecified instead of saying thousands. Valid population sizes sort of depends on the number of clients (a big proxy with millions of users, might want to say a large population is at least 50k, while a small proxy with thousands of users might want to allow smaller populations (for geolocation/caching/etc)).

tfpauly: Fixed!

PR #178, opened 2022-03-07
Clarify security requirements and properties

### LINKED ISSUE #167: Rate-limited tokens: Listing of the different pieces of information and parties
It would be very useful to produce a table that lists in the rows all the different pieces of information, and in column, who learns that information. Not only this would help the comprehension when reading the text, it would also help understanding right away why an attester cannot collude with an issuer or an origin, since basically all the rows would have checkboxes, i.e., a collusion of an attester and an origin would enable to learn most of the pieces of information. It would help for example with Section 9.4.

### LINKED ISSUE #170: Rate-Limited Tokens: Secrecy of public keys?
Section 2 states that 

> A public key chosen by the Client and shared only with the Attester.

> [...] whose public key is not shared with anyone.

This puts a strong emphasis on a notion of secrecy for the public keys, which is (1) unusual, as public keys are (always) assumed to be able to be made “public”, and (2) seems to indicate that the underlying signature scheme needs to have a notion of anonymity. 

If such secrecy of the public key is needed for security of the protocol, it should be emphasized, and the properties needed on the underlying signature scheme should be specified.

chris-wood:
Closes #167.
Closes #170.

### DIFF
--- a/draft-privacypass-rate-limit-tokens.md
+++ b/draft-privacypass-rate-limit-tokens.md
@@ -1010,7 +1010,7 @@ The Client learns the output token. It does not learn the Anonymous Issuer Origi
 Client token requests are constructed such that an Issuer cannot distinguish between
 any two token requests from the same Client and two requests from different Clients.
 We refer to this property as issuance unlinkability. This property is achieved
-by construction. In particular, TokenRequest.request_key and TokenRequest.request_signature
+by the way the tokens are constructed. In particular, TokenRequest.request_key and TokenRequest.request_signature
 are the only value in a TokenRequest that is derived from per-Client information, i.e.,
 the Client Secret.
 
@@ -1030,7 +1030,7 @@ flow, as doing this would allow the Issuer to trivially link two requests to the
 Client.
 
 The token request signature is also required to be unforgeable. Informally, unforgeability
-means that one cannot produce a valid (message, signature) pair for any blinding key without
+means that no entity can produce a valid (message, signature) pair for any blinding key without
 access to the private signing key. Importantly, the means the Attester cannot forge
 signatures on behalf of a given Client in an attempt to learn the origin name.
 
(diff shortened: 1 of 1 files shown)


draft-privacypass-rate-limit-tokens.md
```diff
@@ -982,8 +990,54 @@ mutual authentication between both endpoints. Where appropriate, endpoints
...
+Anonymous Issuer Origin ID or any per-Client information used when creating a token request.
+
+The Client learns the output token. It does not learn the Anonymous Issuer Origin ID.
+
+## Token Request Unlinkability and Unforgeability {#issuance-unlinkability}
+
+Client token requests are constructed such that an Issuer cannot distinguish between
+any two token requests from the same Client and two requests from different Clients.
+We refer to this property as issuance unlinkability. This property is achieved
+by construction. In particular, TokenRequest.request_key and TokenRequest.request_signature
```
tfpauly, line 1013:
```suggestion
by the way the tokens are constructed. In particular, TokenRequest.request_key and TokenRequest.request_signature
```

draft-privacypass-rate-limit-tokens.md
```diff
@@ -982,8 +990,54 @@ mutual authentication between both endpoints. Where appropriate, endpoints
...
+
+This unlinkability property is only intended for requests observed by the Issuer.
+In contrast, the Attester is required to link requests from the same Client together
+for the purposes of enforcing rate limits. This Attester does this by observing the
+Client Key. Importantly, the Client Key is not sent to the Issuer during the issuance
+flow, as doing this would allow the Issuer to trivially link two requests to the same
+Client.
+
+The token request signature is also required to be unforgeable. Informally, unforgeability
+means that one cannot produce a valid (message, signature) pair for any blinding key without
```
tfpauly, line 1033:
```suggestion
means that no entity can produce a valid (message, signature) pair for any blinding key without
```

tfpauly (approved):

dvorak42 (approved):

janaiyengar (approved):

PR #179, opened 2022-03-07
Use RateLimit-Limit header

chris-wood:


tfpauly (changes requested):
I think this will lead to conflating two limits, because `RateLimit-Limit` generically means a rate limit between the originator of the request and the server. That's the Attester and Issuer here. If the attester needs to be rate limited overall to the issuer, then it could use `RateLimit-Limit`, which is separate from the per-encrypted-origin-ID token rate limit.

chris-wood:
As discussed on the call, change things to introduce different "issuer profiles" in the issuer configuration object, and all issuers to change policies through that knob.

PR #180, opened 2022-03-07
Pad the origin name (using the ECH policy)

### LINKED ISSUE #149: Origin name encryption padding
Attesters should not learn anything from encrypted origin names, but the length leaks with the current encryption scheme. We should introduce padding, perhaps cribbing [text from ECH](https://datatracker.ietf.org/doc/html/draft-ietf-tls-esni-13#section-6.1.3) as needed.

chris-wood:
Closes #149.

### DIFF
--- a/draft-privacypass-rate-limit-tokens.md
+++ b/draft-privacypass-rate-limit-tokens.md
@@ -744,7 +744,7 @@ Encrypted Origin Name (`encrypted_origin`) as follows:
    aeadID, and all other values of the TokenRequest structure.
 1. Pad origin_name with N zero bytes, where N = 31 - ((L - 1) % 32) and L is the length
    of origin_name. Denote this padding process as the function `pad`.
-1. Encrypt (seal) origin_name with aad as associated data using context, yielding ciphertext ct.
+1. Encrypt (seal) the padded origin_name with aad as associated data using context, yielding ciphertext ct.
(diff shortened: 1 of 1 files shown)


tfpauly (approved):

dvorak42 (approved):

janaiyengar (approved):

draft-privacypass-rate-limit-tokens.md
```diff
@@ -760,13 +762,13 @@ aad = concat(encode(1, keyID),
...
+ct = context.Seal(aad, pad(origin_name))
 encrypted_origin_name = concat(enc, ct)
 ~~~
 
-Issuers reverse this procedure to recover Origin Name by computing the AAD as described
-above and decrypting encrypted_origin_name with their private key skI, the private key
-corresponding to pkI. In pseudocode, this procedure is as follows:
+Issuers reverse this procedure to recover the (padded) Origin Name by computing the AAD as
+described above and decrypting encrypted_origin_name with their private key skI, the private
+key corresponding to pkI. In pseudocode, this procedure is as follows:
```
tfpauly, line 771:
```suggestion
Issuers reverse this procedure to recover the (padded) Origin Name by computing the AAD as
described above and decrypting encrypted_origin_name with their private key skI (the private
key corresponding to pkI), and stripping off padding bytes. In pseudocode, this procedure
is as follows:
```

PR #188, opened 2022-04-01
Align origin_name with auth scheme, allow for empty name

### LINKED ISSUE #147: Relax rate-limited token challenge requirements
It seems totally reasonable for one to use the rate-limited token type to place an upper bound on tokens minted across origins, so the per-origin requirement seems overly restrictive. Same goes for  the interactive requirement, since that only has implications for the origin. In general, these should both be orthogonal to the type of token protocol.

### LINKED ISSUE #187: Align origin name in rate-limited spec with origin info in auth scheme
We can now support multiple origin names in the origin info. The HPKE encrypted info in the rate-limited spec should be that origin info (the list), not the name.

tfpauly:
Closes #147 
Closes #187

### DIFF
--- a/draft-privacypass-rate-limit-tokens.md
+++ b/draft-privacypass-rate-limit-tokens.md
@@ -306,9 +306,12 @@ and are located at the well-known location /.well-known/token-issuer-directory.
 Clients receive challenges for tokens, as described in {{AUTHSCHEME}}.
 
 For the rate-limited token issuance protocol described in this document,
-the token challenge MUST be interactive and per-origin. That is, the
-TokenChallenge structure MUST contain both the redemption_nonce and
-origin_name fields.
+the name of the origin is sent in an encrypted message from the Client
+to the Issuer. If the TokenChallenge.origin_info field contains a single
+origin name, that origin name is used. If the origin_info field contains
+multiple origin names, the client selects the single origin name that
+presented the challenge. If the origin_info field is empty, the
+encrypted message uses a special value, defined in {{encrypt-origin}}.
 
 The HTTP authentication challenge also SHOULD contain the following
 additional attribute:
@@ -359,8 +362,10 @@ necessary state, as described in this section.
 
 A Client is required to have the following information, derived from a given TokenChallenge:
 
-- Origin Name, a hostname referring to the Origin {{!RFC6454}}. This is
-  the value of TokenChallenge.origin_name.
+- Origin Name, a hostname referring to the Origin {{!RFC6454}}. This is the name
+  of the Origin that issued the token challenge. One or more names can be listed
+  in the TokenChallenge.origin_info field. Rate-limited token issuance relies on the
+  client selecting a single origin name from this list if multiple are present.
 - Token Key, a blind signature public key corresponding to the Issuer
   identified by the TokenChallenge.issuer_name.
 - Origin Name Key, a public key used to encrypt request information corresponding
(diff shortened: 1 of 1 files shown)


draft-privacypass-rate-limit-tokens.md
```diff
@@ -736,7 +741,13 @@ Beyond the key configuration inputs, Clients also require the following inputs d
 in {{request-one}}: `token_key_id`, `blinded_msg`, `request_key`, and `name_key_id`.
 
 Together, these are used to encapsulate Origin Name (`origin_name`) and produce
-Encrypted Origin Name (`encrypted_origin`) as follows:
+Encrypted Origin Name (`encrypted_origin`).
+
+`origin_name` contains the name of the origin that initiated the challenge, as
+taken from the TokenChallenge.origin_info field. If the TokenChallenge.origin_info field
+is empty, `origin_name` is set to a fixed string "EmptyOrigin".
```
chris-wood, line 748:
```suggestion
is empty, `origin_name` is set to the empty string "".
```

Setting it to the empty string seems simplest here?

chris-wood: We probably also need text in the Issuer->Attester section, or elsewhere, describing what to do when this value is received. Maybe we need a section specifically on how issuers should manage or configure origin rate limits?

tfpauly: Yeah I was thinking about adding text for what to do when you receive an empty string. I'll add that.

tfpauly: Added some text for this!

draft-privacypass-rate-limit-tokens.md
```diff
@@ -785,6 +798,10 @@ enc, context = SetupBaseR(enc, skI, "TokenRequest")
 origin_name, error = context.Open(aad, ct)
 ~~~
 
+The resulting value of origin_name is used by the Issuer to determine which rate-limit values
+to send to the Attester for enforcement. If the decrypted origin_name is the empty string "",
+the Issuer applies a cross-origin rate-limit policy, if supported.
```
chris-wood, line 803:
```suggestion
the Issuer applies a cross-origin rate-limit policy, if supported. If the decrypted origin_name
is the empty string "" and the Issuer does not support cross-origin rate limiting, then it MUST
abort the protocol as described in {{request-two}}.
```

draft-privacypass-rate-limit-tokens.md
```diff
@@ -635,7 +640,9 @@ Upon receipt of the forwarded request, the Issuer validates the following condit
 Token Keys and Origin Name Keys held by the Issuer.
 - The TokenRequest.encrypted_origin_name can be decrypted using the
 Issuer's private key (the private key associated with Origin Name Key), and matches
-an Origin Name that is served by the Issuer
+an Origin Name that is served by the Issuer. This name might be the empty string "",
+as described in {{encrypt-origin}}, in which case the Issuer applies a cross-origin
+policy if allowed.
```
chris-wood, line 645:
```suggestion
policy if supported. If a cross-origin policy is not supported, this condition is not met.
```

draft-privacypass-rate-limit-tokens.md
```diff
@@ -306,9 +306,12 @@ and are located at the well-known location /.well-known/token-issuer-directory.
...
 For the rate-limited token issuance protocol described in this document,
-the token challenge MUST be interactive and per-origin. That is, the
-TokenChallenge structure MUST contain both the redemption_nonce and
-origin_name fields.
+the name of the origin is sent in an encrypted message from the Client
+to the Issuer. If the TokenChallenge.origin_info field contains a single
+origin name, that origin name is used. If the origin_info field contains
+multiple origin names, the client selects the single origin name that
+presented the challenge. If the origin_info field is empty, the
+encrypted message uses a special value, defined in {{encrypt-origin}}.
```
chris-wood, line 314:
```suggestion
encrypted message is the empty string "".
```
(We also use the special value "" in #request-two, so I figure it's simplest to just inline here.)

chris-wood (approved):

dvorak42 (approved):

PR #189, opened 2022-04-01
Specify requirements to keep origin names private from attesters

### LINKED ISSUE #182: Token limits should not be unique per-origin
Issuers should ensure that the token limit returned is not unique to a single origin and thus leak that information to the attester.

### LINKED ISSUE #184: Clarify that the Issuer needs to serve a sufficient number of Origins
From the list:

>>  Are you assuming that the client enforces some limit on the number of distinct Issuers, as discussed in other Privacy Pass drafts?  If any origin operates its own private Issuer, the Attester ends up learning that the user visited that origin, violating the separation goal.  Do you intend to prevent this?
> 
> Yes, a 1:1 relationship between Issuer and Origin would indeed violate the goal of separating information from the Attester. It still doesn’t allow the Attester to know the origin state, but it does tell the Attester more about the origin identity than we’d want.
> 
> I think we don’t clearly state a requirement that the Client and/or Attester need to collectively check that there are a sufficient number of origins served by each issuer (conversely, that there aren’t too many issuers). As we’re considering Attester deployments, it would be a requirement when onboarding an Issuer to ensure that there is a sufficient set of origins. We can open an issue to add this text.

tfpauly:
Closes #184
Closes #182

### DIFF
--- a/draft-privacypass-rate-limit-tokens.md
+++ b/draft-privacypass-rate-limit-tokens.md
@@ -1090,7 +1090,7 @@ that an Attester can infer which Origin is accessed by observing the rate limit.
 can be mitigated by having many Origins share the same rate-limit value.
 
 Some deployments MAY choose to relax these requirements, such as in cases where the
-origins being accessed are ubiquitous or do not represent user interaction.
+origins being accessed are ubiquitous or do not correspond to user-specific behavior.
 
 ## Collusion Among Different Entities {#collusion}
 
--- a/draft-privacypass-rate-limit-tokens.md
+++ b/draft-privacypass-rate-limit-tokens.md
@@ -1081,7 +1081,7 @@ encrypted in the request between the Client and the Issuer, so the Attester cann
 directly learn the value. However, in order to prevent the Attester from inferring the
 value, additional constraints need to be added:
 
-- Each Issuer SHOULD serve tokens a large number of Origins. A one-to-one relationship
+- Each Issuer SHOULD serve tokens to a large number of Origins. A one-to-one relationship
 between Origin and Issuer would allow an Attester to infer which Origin is accessed
 simply by observing the Issuer identity.
 
--- a/draft-privacypass-rate-limit-tokens.md
+++ b/draft-privacypass-rate-limit-tokens.md
@@ -1086,7 +1086,7 @@ between Origin and Issuer would allow an Attester to infer which Origin is acces
 simply by observing the Issuer identity.
 
 - Issuers SHOULD NOT return rate-limit values that are specific to Origins, such
-that an Attester can recognize an Origin simply by observing the rate limit. This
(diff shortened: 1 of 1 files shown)


draft-privacypass-rate-limit-tokens.md
```diff
@@ -1071,6 +1073,26 @@ for the Issuer. The Attester needs to support key rotation, but ought to disallo
...
 
+## Origin Identification
+
+As stated in {{properties}}, the design of this protocol is such that Attesters cannot
+learn the identity of origins that Clients are accessing. The Origin Name itself is
+encrypted in the request between the Client and the Issuer, so the Attester cannot
+directly learn the value. However, in order to prevent the Attester from inferring the
+value, additional constraints need to be added:
+
+- Each Issuer SHOULD serve tokens a large number of Origins. A one-to-one relationship
```
chris-wood, line 1084:
```suggestion
- Each Issuer SHOULD serve tokens to a large number of Origins. A one-to-one relationship
```

draft-privacypass-rate-limit-tokens.md
```diff
@@ -1071,6 +1073,26 @@ for the Issuer. The Attester needs to support key rotation, but ought to disallo
...
+between Origin and Issuer would allow an Attester to infer which Origin is accessed
+simply by observing the Issuer identity.
+
+- Issuers SHOULD NOT return rate-limit values that are specific to Origins, such
+that an Attester can recognize an Origin simply by observing the rate limit. This
```
chris-wood, line 1089:
```suggestion
that an Attester can infer which Origin is accessed by observing the rate limit. This
```

draft-privacypass-rate-limit-tokens.md
```diff
@@ -1071,6 +1073,26 @@ for the Issuer. The Attester needs to support key rotation, but ought to disallo
...
+can be mitigated by having many Origins share the same rate-limit value, or by introducing
+variability into return the rate-limit value.
```
chris-wood, line 1091:
```suggestion
can be mitigated by having many Origins share the same rate-limit value.
```
This is akin to randomized padding which doesn't work well over repeated samples, so I think we ought to keep just the simple thing.

draft-privacypass-rate-limit-tokens.md
```diff
@@ -1071,6 +1073,26 @@ for the Issuer. The Attester needs to support key rotation, but ought to disallo
...
+
+Some deployments MAY choose to relax these requirements, such as in cases where the
+origins being accessed are ubiquitous or do not represent user interaction.
```
chris-wood, line 1094:
```suggestion
origins being accessed are ubiquitous or do not correspond to user-specific behavior.
```

chris-wood (approved):

dvorak42 (approved):

PR #204, opened 2022-05-24
Sign the entire request wire format

### LINKED ISSUE #202: TokenRequest signature should include everything on the wire
The signature is currently:

```
struct {
   uint16_t token_type = 0x0003;
   uint8_t token_key_id;
   uint8_t issuer_encap_key_id[32];
   uint8_t encrypted_token_request<1..2^16-1>;
   uint8_t request_signature[Nsig]; // signature over concat(token_type, token_key_id, issuer_encap_key_id, enc_token_req without length prefix)
} TokenRequest;
```

But we should do:

```
struct {
   uint16_t token_type = 0x0003;
   uint8_t token_key_id;
   uint8_t issuer_encap_key_id[32];
   uint8_t encrypted_token_request<1..2^16-1>;
   uint8_t request_signature[Nsig]; // sign everything above verbatim
} TokenRequest;
```

It's just good practice to sign everything and it costs us nothing to do so.

chris-wood:
Turns out this doesn't affect the test vectors after all since we don't have vectors that cover the request signature path. 

Closes #202.

### DIFF
--- a/draft-privacypass-rate-limit-tokens.md
+++ b/draft-privacypass-rate-limit-tokens.md
@@ -1003,6 +1003,7 @@ In pseudocode, this is as follows:
 context = concat(token_type,
                  token_key_id,
                  issuer_encap_key_id,
+                 encode(2, len(encrypted_token_request)),
                  encrypted_token_request)
 request_signature = BKS-BlindKeySign(sk_sign, sk_blind, context)
 ~~~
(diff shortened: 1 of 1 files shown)


tfpauly (approved):

draft-privacypass-rate-limit-tokens.md
```diff
@@ -1003,6 +1003,7 @@ In pseudocode, this is as follows:
 context = concat(token_type,
```
dvorak42, line 1003: nit: Maybe update the "Client produce signatures..." to just explicitly call out signing the entire request, rather than listing the individual bits?

dvorak42 (approved):

draft-privacypass-rate-limit-tokens.md
chris-wood: Good idea -- done!

PR #220, opened 2022-08-08
Address token<>origin binding with per-Origin Token Keys

### LINKED ISSUE #217: Rate-Limited Tokens: Exclusive Key Ownership
**Summary**: A client may be able to provide different public keys to the Attester and Issuer, resulting in an incorrect computation of the OPRF.

**Details**:

According to my best reading of the draft, the client generates a blinded key `request_key` and uses it to generate a signature on the encrypted token request.  The `request_key` is sent to the Issuer inside the encrypted `InnerTokenRequest`. Let me call this copy of the `request_key` _rk1_. The client also communicates this `request_key` and the corresponding `request_blind` to the Attester via a request header. Let me call this copy _rk2_. The Attester and Issuer both verify the signature on the encrypted token request; the Attester also verifies the correctness of `request_blind` with respect to _rk2_ and the client long-term public key. 

The question I had: can the client specify a different `request_key` in the two places? If so, the OPRF computation will no longer be correct, which may allow you to get exceed the token rate limit. For this to work, you would need two different public keys such that the signature verifies correctly under both of them.

Preventing such a construction is not a standard property of signature algorithms. The closest I could find was in this paper:

The Provable Security of Ed25519: Theory and Practice
Jacqueline Brendel, Cas Cremers, Dennis Jackson, and Mang Zhao
https://eprint.iacr.org/2020/823 (also published at S&P 2021)

There they define a property called Malicious Strong Universal Exclusive Ownership (M-S-UEO) that says that the adversary cannot create two public keys _k, k'_ and two messages _m, m'_ such that Verify(m,k) = Verify(m,k') = True. This property is sufficient to prevent the above attack, though not strictly necessary, since in our case we also need for m = m'. 

The paper proves that EdDSA is resilient to such attacks **only if** there is a check that avoids the small subgroup on Curve25519. Unfortunately, as the paper points out, RFC8032 does not specify such a check, though, e.g., libsodium implements it. Without the check, a small subgroup attack can be used for a relatively straightforward attack on M-S-UEO (as well as other key substitution attacks discussed in the paper). I have not checked whether this attack would work end-to-end in rate-limited tokens, but I suspect that subgroup confinement gives you a lot of attack freedom. 

The other issue is that ECDSA is almost certainly not resilient to the kind of key substitution we are looking at here, since it doesn't bind the public key to the signature the way EdDSA does, and there are known key substitution attacks against DSA. Again, I have not checked whether an end-to-end attack would work, but I suspect it might. 

**Solution**:

I think this attack can be addressed by making sure that the Attester and Issuer are using the same public key. A simple way to accomplish this would be to have the attester communicate its copy of `request_key` to the Issuer and then having the Issuer verify that the key inside the encrypted `InnerTokenRequest` matches the one from the Attester. A nice way to do this would be to include a cleartext `request_key` inside the `TokenRequest` structure and perhaps also including it in the signature.

### LINKED ISSUE #218: Rate-Limited Tokens: origin binding of the blind signature
What is preventing a client from going to one website while using the rate-limiting policy of another site? More specifically, suppose Alice visits foo.com and gets a challenge nonce from it. She then runs the issuance protocol, but specifies bar.com as the origin in the encrypted token request. As I understand it, as long as foo.com and bar.com both trust the same issuer, this would work because the origin only sees the (blind) signature from the issuer. 

Among many potential problems resulting from this lack of binding is that bar.com may have a much higher rate limit than foo.com. 

@chris-wood suggests that perhaps a partially-blind signature could be used to fix this problem, where the issuer includes the origin as the public input to the partially-blind sig.

chris-wood:
Closes #218
Closes #217

### DIFF
--- a/draft-privacypass-rate-limit-tokens.md
+++ b/draft-privacypass-rate-limit-tokens.md
@@ -459,7 +459,8 @@ As an example, the Issuer's JSON directory could look like:
 Issuers MUST support at least one Token Key per origin. Issuers MAY support
 multiple Token Key values for the same Origin in order to support rotation.
 Origin configuration for Issuers is out of scope for this document, and so
-the mechanism by which Origins obtain their Token Key value is not specified.
+the mechanism by which Origins obtain their Token Key value is not specified
+here.
 
 Issuer directory resources have the media type "application/json"
 and are located at the well-known location /.well-known/token-issuer-directory.
--- a/draft-privacypass-rate-limit-tokens.md
+++ b/draft-privacypass-rate-limit-tokens.md
@@ -698,14 +698,15 @@ blinded_msg, blind_inv = rsabssa_blind(pkI, token_input)
 The Client then uses Client Key to generate its one-time-use request public
 key `request_key` and blind `request_blind` as described in {{client-anon-issuer-origin-id}}.
 
-The Client then computes `token_key_id` as SHA256(public_key), where public_key is a DER-encoded
+The Client then computes `token_key_id` as the least significant byte of the Token Key
+ID, where the Token Key ID is generated as SHA256(public_key) and public_key is a DER-encoded
 SubjectPublicKeyInfo object carrying Token Key. The Client then constructs a InnerTokenRequest
 value, denoted `origin_token_request`, combining `token_key_id`, `blinded_msg`, and a padded
 representation of the origin name as follows:
 
 ~~~
 struct {
-  uint8_t token_key_id[32];
+  uint8_t token_key_id;
   uint8_t blinded_msg[Nk];
   uint8_t padded_origin_name<0..2^16-1>;
 } InnerTokenRequest;
@@ -840,10 +841,10 @@ If any of these conditions is not met, the Issuer MUST return an HTTP 400 error 
 which will forward the error to the client.
 
 The Issuer determines the correct Issuer Key by using the decrypted Origin Name and
-InnerTokenRequest.token_key_id values. If there is no Token Key whose key ID matches
-InnerTokenRequest.token_key_id and whose Origin matches the Origin Name, the Issuer
-MUST return an HTTP 401 error to the Attester, which will forward the error to the
-client. The Attester learns that the client's view of the Origin key was invalid in the process.
+InnerTokenRequest.token_key_id values. If there is no Token Key whose truncated key ID matches
+InnerTokenRequest.token_key_id, the Issuer MUST return an HTTP 401 error to Attester, which will
+forward the error to the client. The Attester learns that the client's view of the Origin key
+was invalid in the process.
 
 ## Issuer-to-Attester Response {#response-one}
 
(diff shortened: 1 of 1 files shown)


draft-privacypass-rate-limit-tokens.md
```diff
@@ -859,9 +839,9 @@ policy if supported. If a cross-origin policy is not supported, this condition i
 If any of these conditions is not met, the Issuer MUST return an HTTP 400 error to the Attester,
 which will forward the error to the client.
 
-The Issuer determines the correct Issuer Key by using the decrypted Origin Name value and
-TokenRequest.token_key_id. If there is no Token Key whose truncated key ID matches
-TokenRequest.token_key_id, the Issuer MUST return an HTTP 401 error to Attester, which will
+The Issuer determines the correct Issuer Key by using the decrypted Origin Name and
+InnerTokenRequest.token_key_id values. If there is no Token Key whose truncated key ID matches
```
dvorak42, line 843: Probably want to explicitly call out that the token key has to match the key ID and origin name, to prevent cross-origin key usage.

"If there is no Token Key whose truncated key ID matches InnerTokenRequest.token_key_id and whose origin matches Origin Name, ..."

draft-privacypass-rate-limit-tokens.md
```diff
@@ -1460,14 +1441,14 @@ identity (as known to the Attester) to the Origin, especially if repeated over m
 
 ## Token Key Management
 
-Issuers SHOULD generate a new (Token Key, Issuer Origin Secret) regularly, and
```
dvorak42, line 1463: Should this pick up a line about why token key reuse across origins would be a bad idea?

dvorak42 (approved):

draft-privacypass-rate-limit-tokens.md
```diff
@@ -715,13 +697,16 @@ blinded_msg, blind_inv = rsabssa_blind(pkI, token_input)
 The Client then uses Client Key to generate its one-time-use request public
 key `request_key` and blind `request_blind` as described in {{client-anon-issuer-origin-id}}.
 
-The Client then constructs a InnerTokenRequest value, denoted `origin_token_request`,
-combining `blinded_msg`,  `request_key`, and the origin name as follows:
+The Client then computes `token_key_id` as the least significant byte of the Token Key
```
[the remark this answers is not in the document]
chris-wood: Good point! We can simplify this and make it 32 bytes.

draft-privacypass-rate-limit-tokens.md
chris-wood: Yep, added.

tfpauly (approved):

draft-privacypass-rate-limit-tokens.md
```diff
@@ -715,13 +698,15 @@ blinded_msg, blind_inv = rsabssa_blind(pkI, token_input)
...
-The Client then constructs a InnerTokenRequest value, denoted `origin_token_request`,
-combining `blinded_msg`,  `request_key`, and the origin name as follows:
+The Client then computes `token_key_id` as SHA256(public_key), where public_key is a DER-encoded
+SubjectPublicKeyInfo object carrying Token Key. The Client then constructs a InnerTokenRequest
+value, denoted `origin_token_request`, combining `token_key_id`, `blinded_msg`, and a padded
+representation of the origin name as follows:
 
 ~~~
 struct {
+  uint8_t token_key_id[32];
```
tfpauly, line 708: What about removing the key ID altogether? It will prevent any correlation and key targeting attack concerns.

It does mean that during key rotation, you will have failures. But since we say you shouldn't rely on these tokens, that's probably OK?

tfpauly:
@chris-wood I think we also need to update this example in section 7.1.2:

~~~
context = concat(token_type,
                 token_key_id,
                 issuer_encap_key_id,
                 encode(2, len(encrypted_token_request)),
                 encrypted_token_request)
request_signature = BKS-BlindKeySign(sk_sign, sk_blind, context)
~~~

chris-wood:
@tfpauly fixed -- good catch!

PR #223, opened 2022-08-12
Use the key blinding contexts (and keep them separate)

chris-wood:
cc @FredericJacobs 

See the corresponding change [here](https://github.com/cloudflare/pat-go/pull/18/commits/36d6705816141f305e7cba6425639b502d50bad8).

### DIFF
--- a/draft-privacypass-rate-limit-tokens.md
+++ b/draft-privacypass-rate-limit-tokens.md
@@ -1122,18 +1122,19 @@ and {{attester-output-anon-issuer-origin-id}} describes the final Attester step 
 computing the client-origin index.
 
 The index computation is based on a signature scheme with key blinding and unblinding
-support, denoted BKS, as described in {{KEYBLINDING}}. Such a scheme has the following
-functions:
+support, denoted BKS, as described in {{Section 3 of KEYBLINDING}}. Such a scheme has the
+following functions:
 
 - BKS-KeyGen(): Generate a random private and public key pair (sk, pk).
 - BKS-BlindKeyGen(): Generate a random blinding key bk.
-- BKS-BlindPublicKey(pk, bk): Produce a blinded public key based on the input public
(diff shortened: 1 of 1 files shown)


tfpauly (approved):

draft-privacypass-rate-limit-tokens.md
```diff
@@ -1166,6 +1167,7 @@ In pseudocode, this is as follows:
 
 ~~~
 sk_blind = BKS-BlindKeyGen()
+ctx = concat(encode(2, token_type)), "ClientBlind")
 blinded_key = BKS-BlindPublicKey(pk_sign, sk_blind)
```
tfpauly, line 1171:
```suggestion
blinded_key = BKS-BlindPublicKey(pk_sign, sk_blind, ctx)
```

draft-privacypass-rate-limit-tokens.md
```diff
@@ -1166,6 +1167,7 @@ In pseudocode, this is as follows:
 
 ~~~
 sk_blind = BKS-BlindKeyGen()
+ctx = concat(encode(2, token_type)), "ClientBlind")
```
FredericJacobs, line 1170: Is there something else we want to bind in the context?

Not thinking of anything in particular but just the token type and a string does not seem like it's binding a lot.

Maybe something to take another look at in the pats-analysis?

FredericJacobs (approved):
Looks good to me overall. I wonder if the binding in the context is sufficient or if other things should be added.

draft-privacypass-rate-limit-tokens.md
chris-wood: So I think we need to this context to say "this is client blinding for the rate limited privacy pass protocol," and I think what we have suffices for that. But I'm open to adding more if we think it's needed.

PR #231, opened 2022-12-09
Remove SVCB reference

### LINKED ISSUE #229: draft-pauly-httpbis-alias-proxy-status: do not include SVCB/HTTPS records
Per draft-ietf-dnsop-svcb-https-11 section 3.2 clients MUST send the final TargetName
to proxies:
https://datatracker.ietf.org/doc/html/draft-ietf-dnsop-svcb-https-11#section-3.2

Proxies themselves shouldn't be doing SVCB/HTTPS RR lookups to get A/AAAA records.
As such, draft-pauly-httpbis-alias-proxy-status only needs to look at CNAME chains.

(A question might be whether any special handling is needed for DNAME?  That might be its own issue.)

tfpauly:
Closes #229


Smithad499 (commented):
3e8c73c91f7a2d073117ebe33b1281b9a2d75a6f

PR #244, opened 2024-02-29
Make proxies a sub-dictionary

tfpauly:
Address feedback from IETF 118 to allow more information for proxies, and be explicit on proxy protocol

### DIFF
--- a/draft-pauly-intarea-proxy-config-pvd.md
+++ b/draft-pauly-intarea-proxy-config-pvd.md
@@ -102,7 +102,7 @@ In order to fetch PvD Additional Information associated with a proxy, a client
 issues an HTTP GET request for the well-known PvD URI (".well-known/pvd") {{PVDDATA}}
 and the host authority of the proxy. This is applicable for both proxies that are identified
 by a host and port only (such as SOCKS proxies and HTTP CONNECT proxies) and proxies
-that are identified by a URI.
+that are identified by a URI or URI template.
 
 For example, a client would issue the following request for the PvD associated
 with "https://proxy.example.org/masque{?target_host,target_port}":
@@ -163,7 +163,7 @@ use the `:protocol` pseudo-header), the `protocol` value will match the Upgrade 
 
 | Proxy Protocol | Reference |
 | --- | --- |
-| socks | {{!SOCKS=RFC1928}} |
+| socks5 | {{!SOCKSv5=RFC1928}} |
 | connect | {{Section 9.3.6 of HTTP}} |
 | connect-udp | {{CONNECTUDP}} |
 | connect-ip | {{CONNECTIP}} |
--- a/draft-pauly-intarea-proxy-config-pvd.md
+++ b/draft-pauly-intarea-proxy-config-pvd.md
@@ -36,8 +36,8 @@ and a list of DNS zones that are accessible via a proxy.
 HTTP proxies that use the CONNECT method {{Section 9.3.6 of !HTTP=RFC9110}}
 (often referred to as "forward" proxies) allow clients to open connections to
 hosts via a proxy. These typically allow for TCP stream proxying, but can also support
-UDP proxying {{!CONNECTUDP=RFC9298}} and IP packet proxying
-{{!CONNECTIP=RFC9484}}. Such proxies are not just defined as
+UDP proxying {{!CONNECT-UDP=RFC9298}} and IP packet proxying
+{{!CONNECT-IP=RFC9484}}. Such proxies are not just defined as
 hostnames and ports, but can use URI templates {{!URITEMPLATE=RFC6570}}.
 
 In order to make use of multiple related proxies, clients need a way to understand
@@ -165,9 +165,9 @@ use the `:protocol` pseudo-header), the `protocol` value will match the Upgrade 
 | --- | --- |
 | socks5 | {{!SOCKSv5=RFC1928}} |
 | connect | {{Section 9.3.6 of HTTP}} |
-| connect-udp | {{CONNECTUDP}} |
-| connect-ip | {{CONNECTIP}} |
-| connect-tcp | {{!CONNECTTCP=I-D.ietf-httpbis-connect-tcp}} |
+| connect-udp | {{CONNECT-UDP}} |
+| connect-ip | {{CONNECT-IP}} |
+| connect-tcp | {{!CONNECT-TCP=I-D.ietf-httpbis-connect-tcp}} |
 
 The value of "uri" can be either a URI or a URI template, depending on the
 proxy protocol.
@@ -182,8 +182,8 @@ method, with the assumption that the client already is aware of one proxy.
 Many historical methods of configuring a proxy only allow configuring
 a single FQDN hostname for the proxy. A client can attempt to fetch the
 PvD information from the well-known URI to learn the list of complete
-URIs that support non-default protocols, such as {{CONNECTUDP}} and
-{{CONNECTIP}}.
+URIs that support non-default protocols, such as {{CONNECT-UDP}} and
+{{CONNECT-IP}}.
 
 ## Example
 
@@ -223,7 +223,7 @@ content-length = 222
 }
 ~~~
 
-The client would learn the URI template of the proxy that supports UDP using {{CONNECTUDP}},
+The client would learn the URI template of the proxy that supports UDP using {{CONNECT-UDP}},
 at "https://proxy.example.org/masque{?target_host,target_port}".
 
 # Split DNS information for proxies {#split-dns}
--- a/draft-pauly-intarea-proxy-config-pvd.md
+++ b/draft-pauly-intarea-proxy-config-pvd.md
@@ -143,7 +143,7 @@ Each proxy is defined by a proxy protocol, a proxy URI (or URI template
 {{!URITEMPLATE=RFC6570}}), along with potentially other keys.
 
 This document defines two mandatory keys for the sub-dictionaries in the
-`proxies` array, `protocol` and `uri`. These keys, defined below, are
+`proxies` array, `protocol` and `proxy`. These keys, defined below, are
 the initial contents of the proxy information key registry
 ({{proxy-info-iana}}). Other optional keys can be added to the dictionary
 to further define or restrict the use of a proxy. Clients that do not
@@ -154,23 +154,24 @@ uses.
 | JSON Key | Description | Type | Example |
 | --- | --- | --- | --- |
 | protocol | The protocol used to communicate with the proxy | String | "connect-udp" |
-| uri | URI or URI template of the proxy | String | "https://proxy.example.org:4443/masque{?target_host,target_port}" |
+| proxy | String containing the URI template or hostname and port of the proxy, depending on the format defined by the protocol | String | "https://proxy.example.org:4443/masque{?target_host,target_port}" |
 
 The values for the `protocol` key are defined in the proxy protocol
-registry ({{proxy-protocol-iana}}). For proxy types that use HTTP Upgrade Tokens (and
-use the `:protocol` pseudo-header), the `protocol` value will match the Upgrade Token
-/ `:protocol` value.
+registry ({{proxy-protocol-iana}}). For consistency, any new proxy types
+that use HTTP Upgrade Tokens (and use the `:protocol` pseudo-header) SHOULD
+define the `protocol` value to match the Upgrade Token / `:protocol` value.
 
-| Proxy Protocol | Reference |
-| --- | --- |
-| socks5 | {{!SOCKSv5=RFC1928}} |
-| connect | {{Section 9.3.6 of HTTP}} |
-| connect-udp | {{CONNECT-UDP}} |
-| connect-ip | {{CONNECT-IP}} |
-| connect-tcp | {{!CONNECT-TCP=I-D.ietf-httpbis-connect-tcp}} |
+| Proxy Protocol | Proxy Location Format | Reference | Notes |
+| --- | --- | --- |
+| socks5 | hostname:port | {{!SOCKSv5=RFC1928}} | |
+| http-connect | hostname:port | {{Section 9.3.6 of HTTP}} | Standard CONNECT method, using unencrypted HTTP to the proxy |
+| https-connect | hostname:port | {{Section 9.3.6 of HTTP}} | Standard CONNECT method, using TLS-protected HTTP to the proxy |
+| connect-udp | URI template | {{CONNECT-UDP}} | |
+| connect-ip | URI template | {{CONNECT-IP}} | |
+| connect-tcp | URI template | {{!CONNECT-TCP=I-D.ietf-httpbis-connect-tcp}} | |
 
-The value of "uri" can be either a URI or a URI template, depending on the
-proxy protocol.
+The value of `proxy` depends on the Proxy Location Format defined by proxy protocol.
+The types defined here either use a hostname and port, or a full URI template.
 
 When a PvD that contains the `proxies` key is fetched from a known proxy
 using the method described in {{proxy-pvd}} the proxies list describes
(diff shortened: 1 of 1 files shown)


tfpauly:
Rendered format is here https://tfpauly.github.io/privacy-proxy/tfp/proxies-sub-dict/draft-pauly-intarea-proxy-config-pvd.html

draft-pauly-intarea-proxy-config-pvd.md
```diff
@@ -102,7 +102,7 @@ In order to fetch PvD Additional Information associated with a proxy, a client
 issues an HTTP GET request for the well-known PvD URI (".well-known/pvd") {{PVDDATA}}
 and the host authority of the proxy. This is applicable for both proxies that are identified
 by a host and port only (such as SOCKS proxies and HTTP CONNECT proxies) and proxies
-that are identified by a URL.
+that are identified by a URI.
```
DavidSchinazi, line 105:
```suggestion
that are identified by a URI template.
```

draft-pauly-intarea-proxy-config-pvd.md
```diff
@@ -137,15 +137,40 @@ match the hostname of the HTTP proxy. The "prefixes" array SHOULD be empty by de
...
-key is registered in {{iana}}.
-
-The kind of proxy is implied by the URI scheme and any template variables.
-For example, since UDP proxying {{CONNECTUDP}} has the URI template variables
-`target_host` and `target_port`, the URI
-"https://proxy.example.org:4443/masque{?target_host,target_port}" implies
-that the proxy supports UDP proxying.
+is an array of dictionaries, where each dictionary in the array defines
+a single proxy that is available as part of the PvD (see {{proxies-key-iana}}).
+Each proxy is defined by a proxy protocol, a proxy URI (or URI template
```
DavidSchinazi, line 142: I don't think we have ones that are just URIs, aren't they all templates? And to be pedantic all URIs are URI templates with zero parameters. So I'd just say URI template everywhere to make this simpler

draft-pauly-intarea-proxy-config-pvd.md
```diff
@@ -137,15 +137,40 @@ match the hostname of the HTTP proxy. The "prefixes" array SHOULD be empty by de
...
+{{!URITEMPLATE=RFC6570}}), along with potentially other keys.
+
+This document defines two mandatory keys for the sub-dictionaries in the
+`proxies` array, `protocol` and `uri`. These keys, defined below, are
+the initial contents of the proxy information key registry
+({{proxy-info-iana}}). Other optional keys can be added to the dictionary
+to further define or restrict the use of a proxy. Clients that do not
+recognize or understand a key in a proxy sub-dictionary MUST ignore the entire
```
DavidSchinazi, line 150: Erroring out on unknown keys means it's impossible to extend the dictionary in a backwards-compatible way. In Babel we solved this by having a notion of [mandatory vs non-mandatory](https://www.rfc-editor.org/rfc/rfc8966.html#section-4.4). For example here one way to do this could be to have a prefix in the key that identifies whether it's mandatory or not, such as `!foo` is mandatory but `?bar` isn't. So if you see an unknown key the first character tells you if you can ignore or need to ignore the entire proxy definition.

draft-pauly-intarea-proxy-config-pvd.md
```diff
@@ -137,15 +137,40 @@ match the hostname of the HTTP proxy. The "prefixes" array SHOULD be empty by de
...
+| uri | URI or URI template of the proxy | String | "https://proxy.example.org:4443/masque{?target_host,target_port}" |
+
+The values for the `protocol` key are defined in the proxy protocol
+registry ({{proxy-protocol-iana}}). For proxy types that use HTTP Upgrade Tokens (and
+use the `:protocol` pseudo-header), the `protocol` value will match the Upgrade Token
+/ `:protocol` value.
+
+| Proxy Protocol | Reference |
+| --- | --- |
+| socks | {{!SOCKS=RFC1928}} |
```
DavidSchinazi, line 166: Since some stacks still support SOCKSv4, and there might be a SOCKSv6, maybe add the version?
```suggestion
| socks5 | {{!SOCKSv5=RFC1928}} |
```

draft-pauly-intarea-proxy-config-pvd.md
```diff
@@ -37,7 +37,7 @@ HTTP proxies that use the CONNECT method {{Section 9.3.6 of !HTTP=RFC9110}}
 (often referred to as "forward" proxies) allow clients to open connections to
 hosts via a proxy. These typically allow for TCP stream proxying, but can also support
 UDP proxying {{!CONNECTUDP=RFC9298}} and IP packet proxying
-{{!CONNECTIP=I-D.ietf-masque-connect-ip}}. Such proxies are not just defined as
+{{!CONNECTIP=RFC9484}}. Such proxies are not just defined as
```
DavidSchinazi, line 40: nit: can we do `CONNECT-IP` instead of `CONNECTIP` to match how the other MASQUE documents reference each other? (same for UDP and TCP)

draft-pauly-intarea-proxy-config-pvd.md
```diff
@@ -185,7 +210,16 @@ content-length = 222
   "identifier": "proxy.example.org.",
   "expires": "2023-06-23T06:00:00Z",
   "prefixes": [],
-  "proxies": ["https://proxy.example.org","https://proxy.example.org/masque{?target_host,target_port}"]
+  "proxies": [
+    {
+      "protocol": "connect",
+      "uri": "https://proxy.example.org"
```
DavidSchinazi, line 216: Unextended HTTP CONNECT doesn't have a scheme, only host and port

draft-pauly-intarea-proxy-config-pvd.md
```diff
@@ -275,14 +309,37 @@ proxies, can only be safely used when fetched over a secure TLS-protected connec
...
 This document registers a new key in the "Additional Information PvD Keys" registry.
 
 JSON Key: proxies
 
-Description: Array of proxy URIs associated with this PvD
+Description: Array of proxy dictionaries associated with this PvD
+
+Type: Array of dictionaries
+
+Example: [ { "protocol": "connect", "uri":"https://proxy.example.com" } ]
```
DavidSchinazi, line 324:
```suggestion
Example: [ {
  "protocol": "connect",
  "uri": "proxy.example.com"
} ]
```

draft-pauly-intarea-proxy-config-pvd.md
```diff
@@ -275,14 +309,37 @@ proxies, can only be safely used when fetched over a secure TLS-protected connec
...
+## New PvD Proxy Information Registry {#proxy-info-iana}
+
+IANA is requested to create a new registry "Proxy Information PvD Keys", within the "Provisioning Domains (PvDs)" registry page.
+This new registry reserves JSON keys for use in sub-dictionaries under the `proxies` key.
+The initial contents of this registry are given in {{proxy-enumeration}}.
+
+The status of a key as mandatory or optional is intentionally not denoted in the table to allow for
+flexibility in future use cases. Any new assignments of keys will be considered as optional for the purpose of the mechanism described in this document.
+
+New assignments in the "Proxy Information PvD Keys" registry will be administered by IANA through Expert Review {{!RFC8126}}. Experts are
```
DavidSchinazi, line 335: Should we define an allowed character set for keys?

draft-pauly-intarea-proxy-config-pvd.md
tfpauly:
```suggestion
that are identified by a URI or URI template.
```

draft-pauly-intarea-proxy-config-pvd.md
tfpauly: I understand that you don't send a scheme in the request, but it seems like the proxy could still be identified with a scheme

draft-pauly-intarea-proxy-config-pvd.md
tfpauly: Would that be something that would be inherited from the JSON string requirements? We didn't have any such requirement for the main registry spelled out I think.

draft-pauly-intarea-proxy-config-pvd.md
tfpauly: Fixed

draft-pauly-intarea-proxy-config-pvd.md
tfpauly: It seems clearer to me as URI or URI templates, since people may not be used to thinking of templates for socks, etc.

draft-pauly-intarea-proxy-config-pvd.md
tfpauly: Yeah I think this is something to discuss / work on — potentially as a follow-up

draft-pauly-intarea-proxy-config-pvd.md
DavidSchinazi: Well, pedantically there's no scheme though. I think the most correct way to do this would be to define "http-connect" and "https-connect" protocols and then have the uri only contain the host and optional port

draft-pauly-intarea-proxy-config-pvd.md
DavidSchinazi: Technically JSON allows any possible character in keys. That said, some languages have additional restrictions (some C libraries will interpret a null-byte as the end of the key for example, or some high-level languages turn keys into methods but then have requirements on characters allowed in method names such as no spaces, ...). I realize that 8801 didn't constrain its set, but perhaps we can do better here? Wanting consistency makes sense though, in. that case SHOULD limit the characters might be enough

draft-pauly-intarea-proxy-config-pvd.md
DavidSchinazi: Fine by me, please file an issue before merging this though so we don't forget

draft-pauly-intarea-proxy-config-pvd.md
DavidSchinazi: Technically SOCKS doesn't use an URI, just a hostname and port. We could use a URI for SOCKS, but then we'd need to define that the URI can only have a host and port, no scheme, path or query

draft-pauly-intarea-proxy-config-pvd.md
DavidSchinazi: Otherwise we could keep the current URI but explain that only http and https are allowed and explain what it means

draft-pauly-intarea-proxy-config-pvd.md
```diff
@@ -275,14 +309,40 @@ proxies, can only be safely used when fetched over a secure TLS-protected connec
...
+
+## New PvD Proxy Protocol Registry {#proxy-protocol-iana}
 
-Type: Array of strings
+IANA is requested to create a new registry "Proxy Protocol PvD Values", within the "Provisioning Domains (PvDs)" registry page.
+This new registry reserves JSON values for the `protocol` key in `proxies` sub-dictionaries.
+The initial contents of this registry are given in {{proxy-enumeration}}.
 
-Example: ["https://proxy.example.com", "https://proxy.example.com/masque{?target_host,target_port}"]
+New assignments in the "Proxy Protocol PvD Values" registry will be administered by IANA through Specification Required {{!RFC8126}}.
```
DavidSchinazi, line 347: Can we just use Expert Review? This registry has infinite space so restricting allocation will only lead to squatting

tfpauly: Since we need a reference document to point to in the table, it felt like a specification required one?

DavidSchinazi: Can't we say that this field is allowed to be empty? That's what many other registries do

DavidSchinazi: e.g. https://www.iana.org/assignments/masque/masque.xhtml

draft-pauly-intarea-proxy-config-pvd.md
```diff
@@ -137,15 +137,40 @@ match the hostname of the HTTP proxy. The "prefixes" array SHOULD be empty by de
...
+| JSON Key | Description | Type | Example |
+| --- | --- | --- | --- |
+| protocol | The protocol used to communicate with the proxy | String | "connect-udp" |
+| uri | URI or URI template of the proxy | String | "https://proxy.example.org:4443/masque{?target_host,target_port}" |
+
+The values for the `protocol` key are defined in the proxy protocol
+registry ({{proxy-protocol-iana}}). For proxy types that use HTTP Upgrade Tokens (and
+use the `:protocol` pseudo-header), the `protocol` value will match the Upgrade Token
+/ `:protocol` value.
+
```
ddragana, line 163: I think we can omit this sentence. The registry defines a name and the reference defines the spec for the protocol. The fact that the header value and the header match is not important.

draft-pauly-intarea-proxy-config-pvd.md
ddragana: Personally, I prefer http-connect and https-connect.
Maybe it would be clearer to use host name and port instead of URI for CNNECT and socks proxies.

draft-pauly-intarea-proxy-config-pvd.md
tfpauly: Okay, adjusted to allow a hostname:port

draft-pauly-intarea-proxy-config-pvd.md
tfpauly: Filed https://github.com/tfpauly/privacy-proxy/issues/245

draft-pauly-intarea-proxy-config-pvd.md
tfpauly: @ddragana updated!

draft-pauly-intarea-proxy-config-pvd.md
tfpauly: Filed https://github.com/tfpauly/privacy-proxy/issues/246

draft-pauly-intarea-proxy-config-pvd.md
tfpauly: Allowing it to be empty now

draft-pauly-intarea-proxy-config-pvd.md
tfpauly: @ddragana I updated the sentence to just explain that this is nice for consistency. It doesn't matter technically, just is less confusing.

tfpauly:
@DavidSchinazi @ddragana please take a look! I'd like to publish before the deadline Monday.

draft-pauly-intarea-proxy-config-pvd.md
```diff
@@ -137,15 +137,41 @@ match the hostname of the HTTP proxy. The "prefixes" array SHOULD be empty by de
...
-key is registered in {{iana}}.
-
-The kind of proxy is implied by the URI scheme and any template variables.
-For example, since UDP proxying {{CONNECTUDP}} has the URI template variables
-`target_host` and `target_port`, the URI
-"https://proxy.example.org:4443/masque{?target_host,target_port}" implies
-that the proxy supports UDP proxying.
+is an array of dictionaries, where each dictionary in the array defines
+a single proxy that is available as part of the PvD (see {{proxies-key-iana}}).
+Each proxy is defined by a proxy protocol, a proxy URI (or URI template
```
ddragana, line 142:
```suggestion
Each proxy is defined by a proxy protocol, a proxy location (i.e. host and port or URI template
```

draft-pauly-intarea-proxy-config-pvd.md
ddragana: Looks good, thanks.

ddragana (approved):

PR #247, opened 2024-03-05
Add support for DNS zone exclusion lists

joshco:
TODO: make new key in PVD spec.  Update accordingly if necessary


tfpauly:
I've opened #248 — let's track the issue there. I think we may want to discuss more if the key goes at the top level, or within a proxy sub-dictionary.

joshco:
Sounds like a good idea

PR #256, opened 2024-07-26
Harmonize SVCB parameters with WKECH

bemasc:
This is more flexible and consistent.


tfpauly:
Nice! We should probably update the examples, etc too

bemasc:
> Nice! We should probably update the examples, etc too

OK, done

Aundras (approved):

PR #258, opened 2024-10-06
JavaScript engine inconsistencies is the biggest problem with PAC files

yaroslavros:
Variations between PAC file parsing by different JavaScript engine is a major issue... Some software even has to carry around full blown JavaScript engine just to parse PAC files as system libraries cannot always be trusted.

Let's call this out.

### DIFF
--- a/draft-ietf-intarea-proxy-config.md
+++ b/draft-ietf-intarea-proxy-config.md
@@ -87,8 +87,8 @@ hostname and port, and do not support configuring a full URI template
 
 The mechanisms defined in this document are intended to offer a standard
 alternative that works for URI-based proxies and avoids dependencies
-on executing Javascript scripts, which can open up security vulnerabilities and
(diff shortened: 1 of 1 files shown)


draft-ietf-intarea-proxy-config.md
```diff
@@ -87,7 +87,8 @@ hostname and port, and do not support configuring a full URI template
 
 The mechanisms defined in this document are intended to offer a standard
 alternative that works for URI-based proxies and avoids dependencies
-on executing Javascript scripts, which can open up security vulnerabilities.
+on executing Javascript scripts, which can open up security vulnerabilities and
+prone to implementation-specific inconsistencies.
```
tfpauly, line 91:
```suggestion
on executing Javascript scripts, which are prone to implementation-specific inconsistencies and can open up security vulnerabilities.
```

tfpauly (approved):

PR #259, opened 2024-10-06
Domain, IP and port based destinations instead of split-DNS

yaroslavros:
Proxies are often selected not only based on zones, but also based on:
- specific hostnames
- IP addresses or subnets
- destination ports

This proposed change replaces zone based split-DNS with more traditional hostname/wildcard based matching and adds IP and port based selection criteria.

While this won't bring full parity with PAC files, this should cover vast majority of practical use cases.


draft-ietf-intarea-proxy-config.md
```diff
@@ -233,38 +233,58 @@ content-length = 222
...
 This document defines two optional keys that for subdictionaries in the `proxies`
 array that are used for split-DNS configuration.
 
 | JSON Key | Optional | Description | Type | Example |
 | --- | --- | --- | --- | --- |
-| matchDomains | Yes | An array of DNS zones or subdomains that can be accessed over this proxy | Array of Strings | [ "example.com" ] |
-| excludedDomains | Yes | An array of DNS zones or subdomains that cannot be accessed over this proxy, which should be more specific domains of entries in the matchDomains array | Array of Strings | [ "public.example.com" ] |
+| matchDomains | Yes | An array of hostnames and wildcard FQDNs that can be accessed over this proxy | Array of Strings | [ "www.example.com", "*.local" ] |
+| excludedDomains | Yes | An array of hostnames and wildcard FQDNs that cannot be accessed over this proxy, which should be more specific domains of entries in the matchDomains array | Array of Strings | [ "exclude.local" ] |
+| matchIPs | Yes | An array of IP addresses and subnets that can be accessed over this proxy | Array of Strings | [ "2001:DB8::1", "192.168.1.0/24" ] |
```
tfpauly, line 252: I wonder if this should be `matchSubnets`, and the specific IP is a special case of that (a `/128` or `/32`)

yaroslavros:
I think matching both subnets and individual IPs is more user-friendly and will make it easier to migrate from PAC files. I've seen PAC files with thousands (yes, really) individual IPs in proxy steering configurations...
Distinguishing between subnets and IPs is trivial and does not create ambiguity for parsers.

However I don't have an overwhelmingly strong opinion on this, so happy to change this to just subnets if you believe that consistency benefits overweight simplicity for users.

draft-ietf-intarea-proxy-config.md
```diff
@@ -233,38 +233,58 @@ content-length = 222
...
+| excludeIPs | Yes | An array of IP addresses and subnets that cannot be accessed over this proxy, which should be more specific addresses or subnets of entries in the matchIP array | Array of Strings | [ "192.168.1.0/25", "192.168.1.254" ] |
+| matchPorts | Yes | An array of TCP or UDP port ranges accessible over this proxy | Array of Strings | [ "80", "443", "1024-65535" ] |
+| excludePorts | Yes | An array of TCP or UDP port ranges that cannot be accessed over this proxy, which should be more specific than entries in the matchPorts array | Array of Strings | [ "3389", "8443" ] |
```
ddragana, line 255: I am not sure about excludePorts as define here. Ports are a limited space.  Therefore you can define everything by only using matchPorts. The parsing will be easier as well.
Maybe excludedPorts should be use only without matchPorts, to define ports that are disallowed, i.e. use one or the other.
We should also define that ports above 65535 is an error, do we want to reject the configuration or just ignore the mistake?

I think we should have a limit of the number of elements in the arrays, or a recommendation that the arrays should not be longer than x elements because the implementation may reject it. Also we should define the behavior in this case, e.g. the config is rejected as a whole or elements after element x are ignored. This is a separate issue though, and I will open a new issue for it.

yaroslavros: Great ideas!
I'll update PR accordingly.

I don't think there is a concrete list size recommendation that will apply reasonably well to all kinds of clients and use cases - but I think it's a good suggestion to put more important entries in the beginning of the list.

ddragana: Short correction: by "the parsing will be easier as well" I meant to say that it will be easier to create a single list instead of going through 2 lists to parse it into a single list or always check 2 lists.

yaroslavros: Absolutely! PR amended.

draft-ietf-intarea-proxy-config.md
```diff
@@ -233,38 +233,63 @@ content-length = 222
...
 Entries listed in `matchDomains` MUST NOT expand the set of domains that a client is
 willing to send to a particular proxy. The list can only narrow the list of domains
 that the client is willing to send through the proxy. For example, if the client
-has a local policy to only send requests for "example.com" to a proxy
+has a local policy to only send requests for "*.example.com" to a proxy
 "proxy.example.com", and the `matchDomains` array contains "internal.example.com" and
 "other.company.com", the client would end up only proxying "internal.example.com"
 through the proxy.
 
+Wildcard prefix (`*.`) is used to differentiate between specific hostnames and zones. Note
```
tfpauly, line 271:
```suggestion
A wildcard prefix (`*.`) is used to indicate matching entire domains or subdomains instead of specific hostnames. Note
```

draft-ietf-intarea-proxy-config.md
```diff
@@ -233,38 +233,63 @@ content-length = 222
...
+that it is used to match multiple levels of sub-domain. For example "*.example.com"
```
tfpauly, line 272:
```suggestion
that this can be used to match multiple levels of subdomains. For example "*.example.com"
```

draft-ietf-intarea-proxy-config.md
```diff
@@ -233,38 +233,63 @@ content-length = 222
...
+matches "internal.example.com" as well as "www.public.example.com".
+
+Entries in `matchIPs` correspond to IP addresses and subnets that are available through the
```
tfpauly, line 275:
```suggestion
Entries in `matchSubnets` correspond to IP addresses and subnets that are available through the
```

draft-ietf-intarea-proxy-config.md
```diff
@@ -233,38 +233,63 @@ content-length = 222
...
+proxy, while entries in `excludedIPs` define IP addresses and subnets that SHOULD NOT be used
```
tfpauly, line 276:
```suggestion
proxy, while entries in `excludeSubnets` define IP addresses and subnets that SHOULD NOT be used
```

draft-ietf-intarea-proxy-config.md
```diff
@@ -233,38 +233,63 @@ content-length = 222
...
+with the proxy. IP address based destination information SHOULD only be used when
```
tfpauly, line 277:
```suggestion
with the proxy. Subnet-based destination information SHOULD only be used when
```

draft-ietf-intarea-proxy-config.md
```diff
@@ -233,38 +233,63 @@ content-length = 222
...
+communicating with destinations defined by an IP address and not a hostname.
```
tfpauly, line 278:
```suggestion
applications are communicating with destinations identified by only an IP address and not a hostname.
```

draft-ietf-intarea-proxy-config.md
```diff
@@ -233,38 +233,63 @@ content-length = 222
...
+
+`ports` in a list of strings that can be used to instruct the client that only specific destination
```
tfpauly, line 280:
```suggestion
`matchPorts` in a list of strings that can be used to instruct the client that only specific destination
```

draft-ietf-intarea-proxy-config.md
```diff
@@ -233,38 +233,63 @@ content-length = 222
...
+TCP or UDP ports are accessible through the proxy. The list may contain individual port numbers 
+(such as "80") or inclusive ranges of ports. For example "1024-2048" matches all ports from 1024 
+to 2048 including the boundaries.
```
tfpauly, line 283:
```suggestion
to 2048, including the 1024 and 1028.
```

draft-ietf-intarea-proxy-config.md
```diff
@@ -233,38 +233,63 @@ content-length = 222
...
+
+Note that clients with limited resources MAY not be able to process and utilize all
+entries of an excessively long list. In the case if the provided list is too long for a given
+client, it SHOULD process as many records from the beginning of `matchDomains`, `matchIPs` and
```
tfpauly, line 287:
```suggestion
client, it SHOULD process as many records from the beginning of `matchDomains`, `matchSubnets` and
```

draft-ietf-intarea-proxy-config.md
```diff
@@ -233,38 +233,63 @@ content-length = 222
...
+`ports` lists. These lists SHOULD be sorted with the most important elements placed at
```
tfpauly, line 288:
```suggestion
`matchPorts` lists. These lists SHOULD be sorted with the most important elements placed at
```

draft-ietf-intarea-proxy-config.md
```diff
@@ -233,38 +233,63 @@ content-length = 222
...
+to 2048, including the 1024 and 1028.
+
+Note that clients with limited resources MAY not be able to process and utilize all
+entries of an excessively long list. In the case if the provided list is too long for a given
+client, it SHOULD process as many records from the beginning of `matchDomains`, `matchSubnets` and
+`matchPorts` lists. These lists SHOULD be sorted with the most important elements placed at
+the start of the list. If a client cannot consume all the entries in `excludedDomains` or
+`excludedIPs` it SHOULD NOT use given proxy configuration to avoid sending traffic that the proxy
+cannot process.
+
```
tfpauly, line 292: I'm a bit concerned about this section so would prefer to remove it from the first cut of adding this support — let's discuss limited clients more!
```suggestion
```

draft-ietf-intarea-proxy-config.md
```diff
@@ -233,38 +233,55 @@ content-length = 222
...
 that the client is willing to send through the proxy. For example, if the client
-has a local policy to only send requests for "example.com" to a proxy
+has a local policy to only send requests for "*.example.com" to a proxy
 "proxy.example.com", and the `matchDomains` array contains "internal.example.com" and
 "other.company.com", the client would end up only proxying "internal.example.com"
 through the proxy.
 
+A wildcard prefix (`*.`) is used to indicate matching entire domains or subdomains instead of specific hostnames. Note
+that this can be used to match multiple levels of subdomains. For example "*.example.com"
+matches "internal.example.com" as well as "www.public.example.com".
```
tfpauly, line 273:
```suggestion
matches "internal.example.com" as well as "www.public.example.com".
Entries that include the wildcard prefix also SHOULD be treated as if they match
an FQDN that only contains the string after the prefix, with no subdomain. So,
an entry in `matchDomains` of "*.example.com" would match the FQDN "example.com",
unless "example.com" were specifically included in `excludeDomains`. This is
done to prevent commonly needing to include both "*.example.com" and "example.com"
in the `matchDomains` list.
```

draft-ietf-intarea-proxy-config.md
```diff
@@ -233,38 +233,61 @@ content-length = 222
...
 When present in a PvD Additional Information dictionary that is retrieved for a proxy
-as described in {{proxy-pvd}}, domains in the `matchDomains` array indicate specific zones
-that are accessible using the proxy. If a hostname is not included in the enumerated
-zones, then a client SHOULD assume that the hostname will not be accessible through the
-proxy. If a hostname is included in the `excludedDomains` array, then the client SHOULD NOT
-access it through the proxy.
+as described in {{proxy-pvd}}, entries in the `matchDomains` array indicate specific hosts
+and zones that are accessible using the proxy. If a hostname is neither matching an specific
+entry nor included in the enumerated zones, then a client SHOULD assume that the hostname
+will not be accessible through the proxy. If a hostname is included in the `excludedDomains`
```
tfpauly, line 260:
```suggestion
will not be accessible through the proxy. If a hostname is included in the `excludeDomains`
```

tfpauly (approved):

draft-ietf-intarea-proxy-config.md
```diff
@@ -238,40 +238,63 @@ content-length = 222
...
+
+Entries in `matchSubnets` correspond to IP addresses and subnets that are available through the
+proxy, while entries in `excludeSubnets` define IP addresses and subnets that SHOULD NOT be used
+with the proxy. Subnet-based destination information SHOULD only be used when
+applications are communicating with destinations identified by only an IP address and not a hostname.
+
+`matchPorts` in a list of strings that can be used to instruct the client that only specific destination
+TCP or UDP ports are accessible through the proxy. The list may contain individual port numbers 
+(such as "80") or inclusive ranges of ports. For example "1024-2048" matches all ports from 1024 
+to 2048, including the 1024 and 1028.
```
tfpauly, line 296:
```suggestion
`matchPorts` in a list of strings that can be used to instruct the client that only specific destination
TCP or UDP ports are accessible through the proxy. The list may contain individual port numbers
(such as "80") or inclusive ranges of ports. For example "1024-2048" matches all ports from 1024
to 2048, including the 1024 and 1028.
```

PR #261, opened 2024-10-12
First take on proxy auth hint

yaroslavros:
Here is my first stab at authentication hints.

Key use case is to allow client proxy selection based on supported authentication methods.

For example, enterprise environment could have an IoT device that is only capable of TLS PSK, a workload that would like to do non-interactive mTLS authentication and an end user device that is happy to do an interactive authentication flow using HTTP methods. All these endpoints could obtain the same information about available proxies and select a proxy that is best suited to their capabilities and preferences.

### DIFF
--- a/draft-ietf-intarea-proxy-config.md
+++ b/draft-ietf-intarea-proxy-config.md
@@ -144,8 +144,8 @@ Each proxy is defined by a proxy protocol, a proxy location (i.e., a hostname an
 {{!URITEMPLATE=RFC6570}}), along with potentially other keys.
 
 This document defines two mandatory keys for the sub-dictionaries in the
-`proxies` array, `protocol` and `proxy`. There are also optional key, including
-`alpn`, and keys for split-DNS defined in {{split-dns}}.
+`proxies` array, `protocol` and `proxy`. There are also optional keys, including
+`alpn`, `auth`, and keys for split-DNS defined in {{split-dns}}.
 Other optional keys can be added to the dictionary
 to further define or restrict the use of a proxy. Clients that do not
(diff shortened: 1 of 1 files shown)


draft-ietf-intarea-proxy-config.md
```diff
@@ -180,6 +181,35 @@ If the `alpn` key is present, it provides a hint for the Application-Layer Proto
...
+| tls-cert | TLS Client Certificate | {{Section 4.4.2 of ?TLSv1.3=RFC8446}} | https-connect, connect-udp, connect-ip, connect-tcp | |
+| tls-psk | Pre-shared key for TLS connection | {{Section 4.2.11 of TLSv1.3}} | https-connect, connect-udp, connect-ip, connect-tcp | |
+| http-proxy | Proxy-Authorization HTTP header in request to the proxy | {{Section 4.4 of ?HTTP-AUTH=RFC7235}} | http-connect, https-connect, connect-udp, connect-ip, connect-tcp | http-schemes |
+
+`socks-methods` MUST NOT be provided if authentication method is not set to `socks5`.
+`socks-methods` contains a list of decimal numbers corresponding to supported
+authentication methods by the SOCKS proxy. Valid values of methods are listed
+in corresponding IANA registry {{?SOCKS-METHODS=IANA.socks-methods}}.
+
+`http-schemes` MUST NOT be provided if authenticaiton method is not set to `http-proxy`.
```
tfpauly, line 209:
```suggestion
`http-schemes` MUST NOT be provided if authentication method is not set to `http-proxy`.
```

yaroslavros:
Closing based on WG consensus.

PR #263, opened 2024-10-14
Update draft-pauly-httpbis-geoip-hint.md

ciaramcmullin:
This PR updates the reasonings behind the IP Geolocation HTTP Client Hint, emphasizing how the hint enhances client control over their location, reduces the need for extensive IP address pools, minimizes dependencies on IP GeoFeeds on the server side, and preserves user experience.

The format for the IP Geo Header remains unchanged in this PR.

### DIFF
--- a/draft-pauly-httpbis-geoip-hint.md
+++ b/draft-pauly-httpbis-geoip-hint.md
@@ -27,33 +27,56 @@ author:
 
 --- abstract
 
-This documents defines an HTTP Client Hint that allows a client to share information
-about its IP Geolocation. This helps ensure that servers have information about location
-that is consistent with what a client expects and what other servers use.
+This document defines an HTTP Client Hint that allows the client to
+share its location, reducing the need for destination servers to map IP
+ranges to locations. By actively providing their location, clients gain
+greater influence and transparency regarding their geolocation.
+Additionally, when IP-hiding technologies like VPNs or proxies are
+used, servers can still deliver geographically relevant experiences
+without relying on large IP address pools to cover regional areas or
+costly third-party GeoFeeds, which can be outdated and inaccurate at
+times.
 
 --- middle
 
 # Introduction {#introduction}
 
-HTTP Client Hints {{!RFC8942}} defines a convention for HTTP headers
-to communicate optional information from clients to servers as hints. This can be done
-conditionally based on if a server claims supports for a particular hint.
+HTTP Client Hints {{!RFC8942}} defines a convention for HTTP headers to
+communicate optional information from clients to servers as hints. This
+can be done conditionally based on if a server claims to support a
+particular hint. A server requests hints by listing them in the
+Accept-CH response header, or via other methods such as ALPS or
+ACCEPT_CH frames.
 
-This document defines a client hint that can be used to send a IP geolocation entry that
-maps to the client's IP address. This location can be used to influence server behavior,
-such as by causing the server to return responses relevant to the client's location.
-The format of the IP geolocation entry is the same as that defined for IP geolocation
-feeds in {{!GEOFEED=RFC8805}}.
+This document defines a client hint that can be used to send a
+geolocation entry based on the client's determined location. This
+location can be used to influence server behavior, such as by causing
+the server to return responses relevant to the client's location. The
+format of the geolocation entry is the same as that defined for IP
+geolocation feeds in {{!GEOFEED=RFC8805}}. It only allows for
+coarse-level location specification.
 
-This header is intended to be used to provide rough geolocation hints to servers that do
-not already have accurate or authoritative mappings for the IP addresses of clients. This
-can be particularly useful for cases where IP geolocation mappings have changed recently,
-or a client is using a VPN or proxy that may not be commonly recognized by servers.
+This header aims to provide rough geolocation hints to servers based on
+the client’s determined location, shifting geolocation from a passive
+IP-based approach to an active client-controlled one. This not only
+allows the client to influence how their location is interpreted, but
+it also reduces the need for extensive IP address pools when clients
+mask their IP addresses through VPNs or proxies. Typically, VPN
+providers need to purchase egress IPs for each region to maintain
+accurate geolocation. With client-determined location, the hint can
+minimize the number of IP addresses needed while still supporting
+location-based services such as weather, local news, and search
+results. In addition, the hint reduces reliance on third-party
+GeoFeeds which often come with limitations such as outdated
+IP-to-location mappings and ongoing maintenance costs.
 
-The mechanism for how a client learns the IP geolocation mapping to send is beyond the
-scope of this document. {{?RFC9092}} defines some mechanisms for discovery, but clients
-can also have other mechanisms (such as coordinating with a VPN or proxy that is assigning
-the client a tunnelled or proxied address) to learn what hint to sent.
+The mechanism by which the client determines its geolocation is beyond
+the scope of this document. However, the geolocation should still be
+derived from the IP address.
+
+This draft doesn’t eliminate the need for IP Geolocation, but adds an
+additional source of information for many use cases. See [Security
+Considerations] for more discussion.
 
 ## Requirements
 
@@ -73,63 +96,99 @@ For example, the header for an entry "192.0.2.5,US,US-AL,Alabaster" would be:
     Sec-CH-IP-Geo = "US,US-AL,Alabaster"
 ~~~
 
-This field also defines a parameter, "feed", that contains the URI of the
-IP geolocation feed that is authoritative for this entry. For example:
-
-~~~
-    Sec-CH-IP-Geo = "SG,SG-01,Singapore"; feed="https://noc.ietf.org/geo/google.csv"
-~~~
-
-Servers that can provide different content based on Geohash hints SHOULD include
-the headers in their "Accept-CH" list.
+Given that the Sec-CH-Geo is a high-entropy client hint, meaning it is
+a client hint that is not in the low-entropy hint table, the server
+MUST explicitly opt-in to receive the Geo Client Hint as defined in
+{{?RFC8942}}. It will not be sent by default and the server MAY
+indicate support for this hint via the Accept-CH header in the
+initial response:
 
 ~~~
     Accept-CH: Sec-CH-IP-Geo
 ~~~
 
-Servers also SHOULD indicate for any cacheable content if the IP geo hints will influence
-the cached content, using the "Vary" header.
+Servers SHOULD indicate for any cacheable content if the geo hints
+will influence the cached content, using the 'Vary' header. This will
+indicate that the server may have used this header as a determining
+factor when choosing a response:
 
 ~~~
     Vary: Sec-CH-IP-Geo
 ~~~
 
+# Client Behavior
+
+If possible, the client SHOULD specify their geolocation. If location
+is not available, the client MAY send a default value or none at all.
+How the default value is determined is outside the scope of this
+document. However, the default value MUST NOT be more precise or
+detailed than what could be inferred from the user’s IP address.
+
+The client MAY include the client hint in requests to the server after
+the server has explicitly opted in to receiving the hint, or if the
+client knows of specific server configurations, such as proxy
+settings, that support including the hint.
+
 # Server Behavior
 
-Upon receiving a IP Geolocation Client Hint, a server can use the information to influence
-its behavior in various ways, such as determining the content of HTTP responses.
+Upon receiving a Geolocation Client Hint, a server can use the
+information to influence its behavior in various ways, such as
+determining the content of HTTP responses.
 
-Many servers have existing IP geolocation feeds that they use to identify client locations.
-Servers can choose to use the hint value in one of several ways:
+Servers can choose to use the hint value in one of several ways,
+including:
 
-- Use the client hint information instead of consulting another geolocation feed.
-- Check the value of the "feed" parameter on the header and determine if it is a trusted feed.
-   If this feed is trusted, but is not the default feed used by the server, the server
-   can choose to prefer the feed indicated by the client.
-- Check the value of the "feed" parameter on the header and fetch a copy of the feed
-   to verify the mapping, if a copy of the feed has not been fetched recently.
-- If the feed indicated in the "feed" parameter is unknown or untrusted, but starts
-   becoming common, the server can flag this feed as one to be manually checked and
-   added, if appropriate. This allows servers to automatically discover when new
-   feeds and services are brought up..
+- Using the client hint information instead of consulting IP-based
+geolocation feeds.
+- Serving content that corresponds to the client’s indicated location,
+including delivering region-specific news, weather forecasts, and
+relevant advertisements.
+- Determining service availability and feature access based on the
+client’s indicated location.
 
-If the server is acting as a forward proxy, such as a CONNECT proxy, it can use the hint
-to determine an appropriate geo-mapped IP address to use for outbound connections, or a
-client subnet to present in the EDNS0 Client Subnet extension for DNS queries {{?RFC6891}}
+The server MUST be able to handle situations where geolocation is
+not provided in a request. Since not all web clients, such as curl,
+will send a Geolocation Client, the server MAY defer to alternative
+methods such as IP-based geolocation feeds to provide said value.
+
+If the server is acting as a forward proxy, such as a CONNECT proxy,
+it can use the hint to determine an appropriate geo-mapped IP address
+to use for outbound connections, or a client subnet to present in the
+EDNS0 Client Subnet extension for DNS queries {{?RFC6891}}
 {{?RFC7871}}.
 
-# Security Considerations {#security}
+# Security Considerations {#sec-considerations}
 
-The use of the IP Geolocation Client Hint MUST use the Sec- header prefix as recommended
-in {{!RFC8942}}.
+The use of the Geolocation Client Hint MUST use the Sec- header
+prefix as recommended in {{!RFC8942}}.
 
-Servers MUST NOT use IP Geolocation Client Hints for making security or access-control decisions,
-as the value can be spoofed by a client. The hint is intended only for use in optimizing behavior.
+Servers MUST NOT use Geolocation Client Hints for making security or
+access-control decisions, as the value is provided by the client with
+no additional authentication. The hint is intended only to be used
+for greater user visibility and say over their geolocation.
 
-The value contained in this hint SHOULD be based only on a IP Geolocation feed value for
-an IP address the client is already presenting to a server. In order to avoid disclosing
-any private information, this value MUST not be based on geolocation of the client determined
-by other means, such as physical latitude and longitude coordinates.
+# Privacy Considerations {#privacy}
+
+Any default value provided in this hint MUST NOT be more specific than
+the information that could be obtained from the client's IP address
+and a well-maintained map of IP ranges to locations. In particular,
+when a privacy technology such as a VPN is in use, the default value
+MUST NOT reveal information about the user's location that would
+otherwise be hidden.
+
+To prevent disclosing private information, this value MUST NOT be
+based on other sources of geolocation data, such as physical latitude
+and longitude coordinates. Providing overly precise location
+information could expose sensitive user information especially when
+combined with other identifiable signals. Furthermore, when a client
+designates a location different from that derived from their IP
+address, the combination of designated location and IP may create a
+unique identifier, increasing the risk of cross-site tracking.
+
+The hint MUST NOT be sent by default or in an always-on manner. It
+should only be included in response to explicit server requests (e.g.,
+via the Accept-CH header) and in contexts where sharing location
+data serves a clear purpose, such as for location-based services.
 
 # IANA Considerations {#iana}
 


draft-pauly-httpbis-geoip-hint.md
```diff
@@ -73,63 +96,99 @@ For example, the header for an entry "192.0.2.5,US,US-AL,Alabaster" would be:
...
-This field also defines a parameter, "feed", that contains the URI of the
-IP geolocation feed that is authoritative for this entry. For example:
-
-~~~
-    Sec-CH-IP-Geo = "SG,SG-01,Singapore"; feed="https://noc.ietf.org/geo/google.csv"
-~~~
-
-Servers that can provide different content based on Geohash hints SHOULD include
-the headers in their "Accept-CH" list.
+Given that the Sec-CH-Geo is a high-entropy client hint, meaning it is
```
djmitche, line 99: The examples here include `IP-` in the header name, but the text does not. Since the draft emphasizes that the client should determine the location from its own IP, maybe it makes sense to keep the header name `Sec-CH-IP-Geo`. I don't have strong opinions, and it's probably too early for long discussion on the topic, but the draft should at least be consistent.

tfpauly:
@ciaramcmullin can you update the authors on the document to add yourself (and/or others)? We probably don't want more than 4 listed, but I'd like you to be credited.

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -27,33 +27,56 @@ author:
 
 --- abstract
 
-This documents defines an HTTP Client Hint that allows a client to share information
-about its IP Geolocation. This helps ensure that servers have information about location
-that is consistent with what a client expects and what other servers use.
+This document defines an HTTP Client Hint that allows the client to
+share its location, reducing the need for destination servers to map IP
```
tfpauly, line 31: Saying "share its location" can sound a bit scary

```suggestion
share information about its network geolocation, reducing the need for destination servers to map IP
```

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -27,33 +27,56 @@ author:
...
+ranges to locations. By actively providing their location, clients gain
```
tfpauly, line 32:
```suggestion
ranges to locations. By actively providing this hint, clients gain
```

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -27,33 +27,56 @@ author:
...
+greater influence and transparency regarding their geolocation.
```
tfpauly, line 33:
```suggestion
greater influence and transparency regarding their perceived geolocation.
```

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -27,33 +27,56 @@ author:
...
+Additionally, when IP-hiding technologies like VPNs or proxies are
+used, servers can still deliver geographically relevant experiences
+without relying on large IP address pools to cover regional areas or
+costly third-party GeoFeeds, which can be outdated and inaccurate at
+times.
```
tfpauly, line 38:
```suggestion
Specifically, when clients use IP-hiding technologies like VPNs or
proxies, servers can still deliver geographically-relevant content
without relying on large pools of IP addresses and geo-IP feeds that
can be outdated and inaccurate.
```

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -27,33 +27,56 @@ author:
...
-
-The mechanism for how a client learns the IP geolocation mapping to send is beyond the
-scope of this document. {{?RFC9092}} defines some mechanisms for discovery, but clients
-can also have other mechanisms (such as coordinating with a VPN or proxy that is assigning
-the client a tunnelled or proxied address) to learn what hint to sent.
+HTTP Client Hints {{!RFC8942}} defines a convention for HTTP headers to
+communicate optional information from clients to servers as hints. This
+can be done conditionally based on if a server claims to support a
+particular hint. A server requests hints by listing them in the
+Accept-CH response header, or via other methods such as ALPS or
```
tfpauly, line 48: If you include ALPS, it would need a reference. But ALPS is not something that's actually adopted anywhere right now from what I know, so I wouldn't reference it at all.

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -27,33 +27,56 @@ author:
...
+ACCEPT_CH frames.
+
+This document defines a client hint that can be used to send a
+geolocation entry based on the client's determined location. This
+location can be used to influence server behavior, such as by causing
+the server to return responses relevant to the client's location. The
+format of the geolocation entry is the same as that defined for IP
```
tfpauly, line 55:
```suggestion
format of the geolocation hint is the same as that defined for IP
```

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -27,33 +27,56 @@ author:
...
+geolocation feeds in {{!GEOFEED=RFC8805}}. It only allows for
+coarse-level location specification.
+
+This header aims to provide rough geolocation hints to servers based on
+the client’s determined location, shifting geolocation from a passive
```
tfpauly, line 60:
```suggestion
the client’s network location, shifting geolocation from a passive
```

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -27,33 +27,56 @@ author:
...
+IP-based approach to an active client-controlled one. This not only
+allows the client to influence how their location is interpreted, but
+it also reduces the need for extensive IP address pools when clients
+mask their IP addresses through VPNs or proxies. Typically, VPN
```
tfpauly, line 64:
```suggestion
mask their IP addresses through VPNs or proxies. Typically, VPN or proxy
```

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -27,33 +27,56 @@ author:
...
+providers need to purchase egress IPs for each region to maintain
+accurate geolocation. With client-determined location, the hint can
```
tfpauly, line 66:
```suggestion
accurate geolocation. With a client-provided location hint, the hint can
```

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -27,33 +27,56 @@ author:
...
+minimize the number of IP addresses needed while still supporting
+location-based services such as weather, local news, and search
```
tfpauly, line 68:
```suggestion
location-specific content such as weather, local news, and search
```

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -27,33 +27,56 @@ author:
...
+results. In addition, the hint reduces reliance on third-party
+GeoFeeds which often come with limitations such as outdated
+IP-to-location mappings and ongoing maintenance costs.
```
tfpauly, line 71:
```suggestion
results. In addition, the hint reduces reliance on geo-IP
feeds that often come with limitations such as outdated
IP-to-location mappings and ongoing maintenance costs.
```

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -27,33 +27,56 @@ author:
...
+
+The mechanism by which the client determines its geolocation is beyond
+the scope of this document. However, the geolocation should still be
+derived from the IP address.
```
tfpauly, line 75: This seems like a problem to be totally out of scope. I'd be tempted to leave a TODO here to explain that safe bounds need to be put on this determination.

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -27,33 +27,56 @@ author:
...
+
+This draft doesn’t eliminate the need for IP Geolocation, but adds an
```
tfpauly, line 77: What do you mean by this? What would it mean to eliminate the need? Do you mean for geo-IP feeds?

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -27,33 +27,56 @@ author:
...
+additional source of information for many use cases. See [Security
+Considerations] for more discussion.
```
tfpauly, line 79:
```suggestion
additional source of information for many use cases. See {{sec-considerations}}
 for more discussion.
```

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -73,63 +96,99 @@ For example, the header for an entry "192.0.2.5,US,US-AL,Alabaster" would be:
...
-- If the feed indicated in the "feed" parameter is unknown or untrusted, but starts
-   becoming common, the server can flag this feed as one to be manually checked and
-   added, if appropriate. This allows servers to automatically discover when new
-   feeds and services are brought up..
-
-If the server is acting as a forward proxy, such as a CONNECT proxy, it can use the hint
-to determine an appropriate geo-mapped IP address to use for outbound connections, or a
-client subnet to present in the EDNS0 Client Subnet extension for DNS queries {{?RFC6891}}
-{{?RFC7871}}.
+If possible, the client SHOULD specify their geolocation. If location
```
tfpauly, line 121: ...based on what? This seems pretty vague.

tfpauly: I would say that if they don't have a way to know their external IP granularity, they SHOULD NOT include any value.

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -73,63 +96,99 @@ For example, the header for an entry "192.0.2.5,US,US-AL,Alabaster" would be:
...
+EDNS0 Client Subnet extension for DNS queries {{?RFC6891}}
+{{?RFC7871}}.
 
-The value contained in this hint SHOULD be based only on a IP Geolocation feed value for
-an IP address the client is already presenting to a server. In order to avoid disclosing
-any private information, this value MUST not be based on geolocation of the client determined
-by other means, such as physical latitude and longitude coordinates.
+# Security Considerations {#sec-considerations}
+
+The use of the Geolocation Client Hint MUST use the Sec- header
```
tfpauly, line 162:
```suggestion
The use of the Geolocation Client Hint MUST use the `Sec-` header
```

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -73,63 +96,99 @@ For example, the header for an entry "192.0.2.5,US,US-AL,Alabaster" would be:
...
+Any default value provided in this hint MUST NOT be more specific than
+the information that could be obtained from the client's IP address
+and a well-maintained map of IP ranges to locations. In particular,
+when a privacy technology such as a VPN is in use, the default value
+MUST NOT reveal information about the user's location that would
+otherwise be hidden.
+
+To prevent disclosing private information, this value MUST NOT be
+based on other sources of geolocation data, such as physical latitude
+and longitude coordinates. Providing overly precise location
```
tfpauly, line 181: Maybe mention GPS here

tfpauly (changes requested):

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -27,33 +27,55 @@ author:
 
 --- abstract
 
-This documents defines an HTTP Client Hint that allows a client to share information
-about its IP Geolocation. This helps ensure that servers have information about location
-that is consistent with what a client expects and what other servers use.
+This document defines an HTTP Client Hint that allows the client to
```
DavidSchinazi, line 30: I think this abstract would be more powerful if we reversed the order to be (1) problem then (2) solution.

So something that paraphrases to "We're doing a lot more IP-hiding to improve privacy these days, but that costs a lot of IPs and also users often get bad geo results so users are sad. We fix this by sending the geo from client to origin. We accomplish that with a client hint"

draft-pauly-httpbis-geoip-hint.md
DavidSchinazi: +1, I would just skip ALPS, mainly because ALPS failed at IETF and we don't want to be associated with that

draft-pauly-httpbis-geoip-hint.md
DavidSchinazi: I think this is going to be the main sticking point, so I would resolve it this week. We can say that the hint MUST be determined by a cooperating server looking up the client's IP in a geoIP database. GPS MUST NOT be used. Otherwise we'll get the same opposition we got last time due to privacy concerns

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -73,63 +95,99 @@ For example, the header for an entry "192.0.2.5,US,US-AL,Alabaster" would be:
...
-~~~
-    Sec-CH-IP-Geo = "SG,SG-01,Singapore"; feed="https://noc.ietf.org/geo/google.csv"
-~~~
-
-Servers that can provide different content based on Geohash hints SHOULD include
-the headers in their "Accept-CH" list.
+Given that the Sec-CH-IP-Geo is a high-entropy client hint, meaning it is
+a client hint that is not in the low-entropy hint table, the server
+MUST explicitly opt-in to receive the Geo Client Hint as defined in
+{{?RFC8942}}. It will not be sent by default and the server MAY
```
DavidSchinazi, line 101: The `?` means informative reference, but we want a normative reference. Those are represented with `!` but that's already done on line 43, so just reference without either `?` nor `!`

```suggestion
{{RFC8942}}. It will not be sent by default and the server MAY
```

draft-pauly-httpbis-geoip-hint.md
DavidSchinazi: Then we'd need to define what granularity means. Probably best to be more strict here, and use the normative text I mentioned above
* MUST determine geo using cooperating server and geoIP DB
* client MUST NOT use GPS
* client MAY send this header to servers

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -73,63 +95,99 @@ For example, the header for an entry "192.0.2.5,US,US-AL,Alabaster" would be:
...
-   becoming common, the server can flag this feed as one to be manually checked and
-   added, if appropriate. This allows servers to automatically discover when new
-   feeds and services are brought up..
-
-If the server is acting as a forward proxy, such as a CONNECT proxy, it can use the hint
-to determine an appropriate geo-mapped IP address to use for outbound connections, or a
-client subnet to present in the EDNS0 Client Subnet extension for DNS queries {{?RFC6891}}
-{{?RFC7871}}.
+If possible, the client SHOULD specify their geolocation. If location
+is not available, the client MAY send a default value or none at all.
```
DavidSchinazi, line 121: I don't see the value of a default value. We should either explain it or remove it

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -73,63 +95,99 @@ For example, the header for an entry "192.0.2.5,US,US-AL,Alabaster" would be:
...
+
+Servers can choose to use the hint value in one of several ways,
+including:
+
+- Using the client hint information instead of consulting IP-based
+geolocation feeds.
+- Serving content that corresponds to the client’s indicated location,
+including delivering region-specific news, weather forecasts, and
+relevant advertisements.
+- Determining service availability and feature access based on the
```
DavidSchinazi, line 145: We definitely don't want this one. That would allow clients to work around laws that limit availability. We should instead say to not do this

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -73,63 +95,99 @@ For example, the header for an entry "192.0.2.5,US,US-AL,Alabaster" would be:
...
+client’s indicated location.
+
+The server MUST be able to handle situations where geolocation is
+not provided in a request. Since not all web clients, such as curl,
```
DavidSchinazi, line 149: No need to single out curl, its author might not be too happy about that

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -73,63 +95,99 @@ For example, the header for an entry "192.0.2.5,US,US-AL,Alabaster" would be:
...
+will send a Geolocation Client, the server MAY defer to alternative
+methods such as IP-based geolocation feeds to provide said value.
+
+If the server is acting as a forward proxy, such as a CONNECT proxy,
+it can use the hint to determine an appropriate geo-mapped IP address
+to use for outbound connections, or a client subnet to present in the
+EDNS0 Client Subnet extension for DNS queries {{?RFC6891}}
```
DavidSchinazi, line 156: This text predates this PR, but seems odd to me. @tfpauly what was the point of this? Do you use it? It seems to be quite backwards to go from geo to IP for client subnet. It even seems harmful because you want your DNS to match your data-plane egress, not your actual physical location

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -73,63 +95,99 @@ For example, the header for an entry "192.0.2.5,US,US-AL,Alabaster" would be:
...
+{{?RFC7871}}.
 
-The value contained in this hint SHOULD be based only on a IP Geolocation feed value for
-an IP address the client is already presenting to a server. In order to avoid disclosing
-any private information, this value MUST not be based on geolocation of the client determined
-by other means, such as physical latitude and longitude coordinates.
+# Security Considerations {#sec-considerations}
+
+The use of the Geolocation Client Hint MUST use the `Sec-` header
```
DavidSchinazi, line 161: This MUST seems redundant. This document specifies a header that uses the Sec- prefix. How could someone use this without it?

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -73,63 +95,99 @@ For example, the header for an entry "192.0.2.5,US,US-AL,Alabaster" would be:
...
+prefix as recommended in {{!RFC8942}}.
+
+Servers MUST NOT use Geolocation Client Hints for making security or
+access-control decisions, as the value is provided by the client with
+no additional authentication. The hint is intended only to be used
+for greater user visibility and say over their geolocation.
```
DavidSchinazi, line 167: I think we should stay away from user visibility and say. The intent here isn't to allow the human user to control this

DavidSchinazi (commented):
Overall this looks good. Comments inline

draft-pauly-httpbis-geoip-hint.md
tfpauly: We can probably remove this. I think it is part of an older version.

draft-pauly-httpbis-geoip-hint.md
ciaramcmullin: Got it removed!

draft-pauly-httpbis-geoip-hint.md
ciaramcmullin: I updated to reflect this. LMK what you think. Thanks!

draft-pauly-httpbis-geoip-hint.md
ciaramcmullin: Done.

draft-pauly-httpbis-geoip-hint.md
ciaramcmullin: I rephrased the introduction to reflect this. LMK what you think!

draft-pauly-httpbis-geoip-hint.md
ciaramcmullin: Ah yes removed. Thanks!

draft-pauly-httpbis-geoip-hint.md
ciaramcmullin: Removed.

draft-pauly-httpbis-geoip-hint.md
ciaramcmullin: Removed this!

draft-pauly-httpbis-geoip-hint.md
ciaramcmullin: I believe this also comes from @tfpauly previous draft.

djmitche (commented):
The updated introduction is a big improvement, and I appreciate the "cooperating server" detail.

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -24,36 +24,70 @@ author:
...
+    ins: D. Mitchell
+    name: Dustin Mitchell
+    organization: Google LLC
+    email: djmitche@gmail.com
 --- abstract
 
-This documents defines an HTTP Client Hint that allows a client to share information
-about its IP Geolocation. This helps ensure that servers have information about location
-that is consistent with what a client expects and what other servers use.
+The rise of IP-hiding technologies like VPNs and proxies helps improve user
```
DavidSchinazi, line 39:
```suggestion
The rise of IP-hiding technologies such as VPNs and proxies helps improve user
```

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -24,36 +24,70 @@ author:
...
+privacy, but it also brings some drawbacks. Maintaining a geographically
+relevant user experience requires large pools of IP addresses, which can
+be costly. Additionally, users often receive inaccurate geolocation
+results because servers rely on geo-IP feeds that can be outdated. To
+address these challenges, we can allow clients to actively send their
+network geolocation directly to the origin server via an HTTP Client
+Hint. This approach will not only enhance geolocation accuracy and reduce IP
+costs, but it also gives clients more say and transparency regarding
```
DavidSchinazi, line 47: Do we really want to keep the part about the client having a say? That takes us into the policy space and I worry that this will complicate discussions by moving them away from the technical realm

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -24,36 +24,70 @@ author:
...
-can be particularly useful for cases where IP geolocation mappings have changed recently,
-or a client is using a VPN or proxy that may not be commonly recognized by servers.
-
-The mechanism for how a client learns the IP geolocation mapping to send is beyond the
-scope of this document. {{?RFC9092}} defines some mechanisms for discovery, but clients
-can also have other mechanisms (such as coordinating with a VPN or proxy that is assigning
-the client a tunnelled or proxied address) to learn what hint to sent.
+HTTP Client Hints {{!RFC8942}} defines a convention for HTTP headers to
+communicate optional information from clients to servers as hints. This
+can be done conditionally based on if a server claims to support a
```
DavidSchinazi, line 56:
```suggestion
can be done conditionally based on whether a server claims to support a
```

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -24,36 +24,70 @@ author:
...
+the client’s network location, shifting geolocation from a passive
+IP-based approach to an active client-controlled one. This not only
+allows the client to influence how their location is interpreted, but
+it also reduces the need for extensive IP address pools when clients
+mask their IP addresses through VPNs or proxies. Typically, VPN or proxy
+providers need to purchase egress IPs for each region to maintain
+accurate geolocation. With a client-provided location hint, the hint can
+minimize the number of IP addresses needed while still supporting
+location-specific content such as weather, local news, and search
+results. In addition, the hint reduces reliance on geo-IP
```
DavidSchinazi, line 78:
```suggestion
results. In addition, the hint reduces most servers' reliance on geo-IP
```

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -24,36 +24,70 @@ author:
...
+feeds that often come with limitations such as outdated
+IP-to-location mappings and ongoing maintenance costs.
+
+The client MUST determine geolocation via a cooperating server
+that performs a geo-IP database lookup of the client's IP address.
+Geolocation MUST NOT be derived from GPS or any method that
+provides more specificity than the information obtainable from the
+client's IP address.
+
+This draft doesn’t eliminate the need for IP-based geolocation, but adds an
```
DavidSchinazi, line 88: I'm not sure I understand this sentence: the clause after the "but" seems unrelated to the one before

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -73,63 +107,91 @@ For example, the header for an entry "192.0.2.5,US,US-AL,Alabaster" would be:
...
-IP geolocation feed that is authoritative for this entry. For example:
-
-~~~
-    Sec-CH-IP-Geo = "SG,SG-01,Singapore"; feed="https://noc.ietf.org/geo/google.csv"
-~~~
-
-Servers that can provide different content based on Geohash hints SHOULD include
-the headers in their "Accept-CH" list.
+Given that the Sec-CH-IP-Geo is a high-entropy client hint, meaning it is
+a client hint that is not in the low-entropy hint table, the server
```
DavidSchinazi, line 111:
```suggestion
Given that the Sec-CH-IP-Geo is a high-entropy client hint (i.e.,
a client hint that is not in the low-entropy hint table), the server
```

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -73,63 +107,91 @@ For example, the header for an entry "192.0.2.5,US,US-AL,Alabaster" would be:
...
+MUST explicitly opt-in to receive the Geo Client Hint as defined in
```
DavidSchinazi, line 112: As written, this MUST tells all servers to opt-in, but that's not what we want.
```suggestion
needs to explicitly opt-in in order to receive the Geo Client Hint as defined in
```

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -24,36 +24,70 @@ author:
...
+location-specific content such as weather, local news, and search
+results. In addition, the hint reduces reliance on geo-IP
+feeds that often come with limitations such as outdated
+IP-to-location mappings and ongoing maintenance costs.
+
+The client MUST determine geolocation via a cooperating server
+that performs a geo-IP database lookup of the client's IP address.
+Geolocation MUST NOT be derived from GPS or any method that
+provides more specificity than the information obtainable from the
+client's IP address.
```
DavidSchinazi, line 86: One of the conventions in RFCs is to not duplicate normative language (MUST/SHOULD/etc). Since we already have the MUST in the Client Behavior section, I would use prose here instead

```suggestion
The client determines geolocation via a cooperating server
that performs a geo-IP database lookup of the client's IP address.
```

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -73,63 +107,91 @@ For example, the header for an entry "192.0.2.5,US,US-AL,Alabaster" would be:
...
-
-Many servers have existing IP geolocation feeds that they use to identify client locations.
-Servers can choose to use the hint value in one of several ways:
+# Client Behavior
+ 
+The client MUST determine geolocation using a cooperating server
+that looks up the client's IP address in a geo-IP database. The client
+MUST NOT use GPS. The client hint value MUST NOT be more precise
+or detailed than what can be inferred from the user’s IP address.
+If the client cannot determine how much information its external
```
DavidSchinazi, line 136: I'm not sure what this requirement means, how do we quantify information?

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -73,63 +107,91 @@ For example, the header for an entry "192.0.2.5,US,US-AL,Alabaster" would be:
...
-- Check the value of the "feed" parameter on the header and determine if it is a trusted feed.
-   If this feed is trusted, but is not the default feed used by the server, the server
-   can choose to prefer the feed indicated by the client.
-- Check the value of the "feed" parameter on the header and fetch a copy of the feed
-   to verify the mapping, if a copy of the feed has not been fetched recently.
-- If the feed indicated in the "feed" parameter is unknown or untrusted, but starts
-   becoming common, the server can flag this feed as one to be manually checked and
-   added, if appropriate. This allows servers to automatically discover when new
-   feeds and services are brought up..
+The client MAY append the client hint header in requests to the
```
DavidSchinazi, line 139:
```suggestion
The client MAY include the client hint header in requests to the
```

draft-pauly-httpbis-geoip-hint.md
DavidSchinazi: Makes sense. Let's still remove it though

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -73,63 +107,91 @@ For example, the header for an entry "192.0.2.5,US,US-AL,Alabaster" would be:
...
+The use of the Geolocation Client Hint MUST use the `Sec-` header
+prefix as recommended in {{!RFC8942}}.
+
+Servers MUST NOT use Geolocation Client Hints for making security or
+access-control decisions, as the value is provided by the client with
+no additional authentication.
+
+# Privacy Considerations {#privacy}
+
+Any default value provided in this hint MUST NOT be more specific than
```
DavidSchinazi, line 175: Since we removed the default value from above, let's remove it from here too

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -73,63 +107,91 @@ For example, the header for an entry "192.0.2.5,US,US-AL,Alabaster" would be:
...
+the information that could be obtained from the client's IP address
+and a well-maintained map of IP ranges to locations. In particular,
+when a privacy technology such as a VPN is in use, the default value
+MUST NOT reveal information about the user's location that would
+otherwise be hidden.
+
+To prevent disclosing private information, this value MUST NOT be
```
DavidSchinazi, line 182: Similar to above, let's use prose here to avoid repeating the normative text form the Client Behavior section

```suggestion
To prevent disclosing private information, this value cannot be
```

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -73,63 +107,91 @@ For example, the header for an entry "192.0.2.5,US,US-AL,Alabaster" would be:
...
+not provided in a request. Since not all web clients will send a 
+Geolocation Client Hint, the server MAY defer to alternative methods 
+such as IP-based geolocation feeds to provide said value.
+
+# Security Considerations {#sec-considerations}
+
+The use of the Geolocation Client Hint MUST use the `Sec-` header
+prefix as recommended in {{!RFC8942}}.
+
+Servers MUST NOT use Geolocation Client Hints for making security or
```
DavidSchinazi, line 169: Should we also mention the country portion here. Perhaps: "Servers that only offer services to clients that appear to be in a given country or administrative region still need to rely on a geoIP database to determine the region for access control purposes, but they can use the client hint to customize the response to where the client claims to be inside that region."

DavidSchinazi (commented):
This is getting really close to ready

draft-pauly-httpbis-geoip-hint.md
ciaramcmullin: @tfpauly suggested this change in a previous comment - Tommy could you elaborate here?

draft-pauly-httpbis-geoip-hint.md
ciaramcmullin: Yeah I think we were trying to say we still need IP-based geolocation and this draft is not trying to get rid of it. With the different updates though I think this sentence isn't necessary anymore. I removed it!

draft-pauly-httpbis-geoip-hint.md
ciaramcmullin: Done.

draft-pauly-httpbis-geoip-hint.md
ciaramcmullin: Added!

draft-pauly-httpbis-geoip-hint.md
ciaramcmullin: I actually think it's important to say this since we are trying to emphasize the shift from passive to active fingerprinting here. WDYT?

draft-pauly-httpbis-geoip-hint.md
ciaramcmullin: I was looking at https://www.rfc-editor.org/rfc/rfc8942.html#name-deployment-and-security-ris and it seems like it may be important to say this based on "Using the "Sec-" prefix signals to servers that the user agent -- and not application content -- generated the values." I'm fine with removing it though. WDYT?

draft-pauly-httpbis-geoip-hint.md
ciaramcmullin: This was removed. Resolving comment!

draft-pauly-httpbis-geoip-hint.md
DavidSchinazi: In that case I'd drop "say" and keep "transparency"

```suggestion
costs, but it also gives clients more transparency regarding
```

draft-pauly-httpbis-geoip-hint.md
DavidSchinazi: Yeah let's remove it. It's important that we choose a header that starts with "Sec-" but once we've done that we don't need to say anything about it

draft-pauly-httpbis-geoip-hint.md
DavidSchinazi: I think that Tommy's comment was before we decided to force the use of a cooperating server. I would just remove this sentence now

DavidSchinazi (approved):
Nice! From my perspective, this is good to go once the last few minor comments are resolved

draft-pauly-httpbis-geoip-hint.md
ciaramcmullin: Ok done!

draft-pauly-httpbis-geoip-hint.md
ciaramcmullin: Removed "say".

DavidSchinazi (approved):

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -24,36 +24,63 @@ author:
...
+The rise of IP-hiding technologies such as VPNs and proxies helps improve user
+privacy, but it also brings some drawbacks. Maintaining a geographically
+relevant user experience requires large pools of IP addresses, which can
+be costly. Additionally, users often receive inaccurate geolocation
+results because servers rely on geo-IP feeds that can be outdated. To
+address these challenges, we can allow clients to actively send their
+network geolocation directly to the origin server via an HTTP Client
+Hint. This approach will not only enhance geolocation accuracy and reduce IP
+costs, but it also gives clients more transparency regarding their perceived
+geolocation.
```
tfpauly, line 48:
```suggestion
Techniques that improve user privacy by hiding original client IP addresses,
such as VPNs and proxies, have faced challenges with server that rely on
IP addresses to determine client location. Maintaining a geographically
relevant user experience requires large pools of IP addresses, which can
be costly. Additionally, users often receive inaccurate geolocation
results because servers rely on geo-IP feeds that can be outdated. To
address these challenges, we can allow clients to actively send their
network geolocation directly to the origin server via an HTTP Client
Hint. This approach will not only enhance geolocation accuracy and reduce IP
costs, but it also gives clients more transparency regarding their perceived
geolocation.
```

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -24,36 +24,64 @@ author:
...
+geolocation feeds in {{!GEOFEED=RFC8805}}. It only allows for
+coarse-level location specification.
+
+This header aims to provide rough geolocation hints to servers based on
+the client’s network location, shifting geolocation from a passive
+IP-based approach to an active client-controlled one. This not only
+allows the client to influence how their location is interpreted, but
+it also reduces the need for extensive IP address pools when clients
+mask their IP addresses through VPNs or proxies. Typically, VPN or proxy
+providers need to purchase egress IPs for each region to maintain
```
tfpauly, line 75:
```suggestion
providers need to manage egress IPs for each region to maintain
```

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -73,63 +101,92 @@ For example, the header for an entry "192.0.2.5,US,US-AL,Alabaster" would be:
...
-its behavior in various ways, such as determining the content of HTTP responses.
-
-Many servers have existing IP geolocation feeds that they use to identify client locations.
-Servers can choose to use the hint value in one of several ways:
+# Client Behavior
+ 
+The client MUST determine geolocation using a cooperating server
+that looks up the client's IP address in a geo-IP database. The client
+MUST NOT use GPS. The client hint value MUST NOT be more precise
+or detailed than what can be inferred from the user’s IP address.
```
tfpauly, line 129:
```suggestion
or detailed than what can be inferred from the user’s IP address.
When the client is routing traffic through a proxy or a VPN, the
IP address used to generate this geolocation hint MUST be an
address that is presented upstream beyond the proxy or VPN
(in other words, the "egress IP address"). The proxy or VPN's
selection of this egress IP address MAY have been based on
the client's original un-proxied IP address, but any hints that
the client presents to servers beyond a proxy or VPN MUST NOT
reveal more geolocation information that would be possible to
determine from looking up information about the egress IP address
itself.
```

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -73,63 +101,102 @@ For example, the header for an entry "192.0.2.5,US,US-AL,Alabaster" would be:
...
-by other means, such as physical latitude and longitude coordinates.
+Upon receiving a Geolocation Client Hint, a server can use the
+information to influence its behavior in various ways, such as
+determining the content of HTTP responses.
+
+Servers can choose to use the hint value in one of several ways,
+including:
+
+- Using the client hint information instead of consulting IP-based
+geolocation feeds.
```
tfpauly, line 156:
```suggestion
geolocation feeds.
- Recognizing a mismatch between the client hint information and the server's
current result from its IP-based geolocation feed as a reason to schedule an
automatic refresh of its geolocation feed information. This can help ensure that
changes to feeds are adopted quickly, improving results for clients that don't
send the client hint.
```

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -73,63 +101,107 @@ For example, the header for an entry "192.0.2.5,US,US-AL,Alabaster" would be:
...
+Geolocation Client Hint, the server MAY defer to alternative methods 
+such as IP-based geolocation feeds to provide said value.
+
+# Security Considerations {#sec-considerations}
+
+Servers MUST NOT use Geolocation Client Hints for security or
+access-control decisions, as the value is provided by the client
+without additional authentication or verification. Servers that
+offer services restricted to clients in a specific country or
+administrative region must rely on geoIP databases to
```
tfpauly, line 177:
```suggestion
administrative region might already rely on geoIP databases to
```

draft-pauly-httpbis-geoip-hint.md
```diff
@@ -73,63 +101,107 @@ For example, the header for an entry "192.0.2.5,US,US-AL,Alabaster" would be:
...
+NOT reveal information about the user's location that would
+otherwise be hidden.
+
+To prevent disclosing private information, this value cannot be
+based on other sources of geolocation data, such as GPS or physical 
+latitude and longitude coordinates. Providing overly precise location
+information could expose sensitive user information especially when
+combined with other identifiable signals. Furthermore, when a client
+designates a location different from that derived from their IP
+address, the combination of designated location and IP may create a
```
tfpauly, line 198:
```suggestion
address, the combination of designated location and IP can create a
```

tfpauly (approved):

tfpauly:
Looks like a good checkpoint, thanks all!

Issue #255, opened 2024-07-24
Rethink names matchDomains / excludedDomains


tfpauly:
match - exclude
matched - excluded
etc.

tfpauly:
Closed by #264

PR #264, opened 2024-10-18
Clarify that exclusions can be specified without any match domains

### LINKED ISSUE #254: Clarify that exclusions can be specified without any match domains
For cases where you want to include everything by default, except one name, we can just have excluded domains

ddragana:
Change excludedDomains into excludeDomains to match matchDomains.

This should resolve #254 and #255 .

### DIFF
--- a/draft-ietf-intarea-proxy-config.md
+++ b/draft-ietf-intarea-proxy-config.md
@@ -249,15 +249,15 @@ array that are used for split-DNS configuration.
 | JSON Key | Optional | Description | Type | Example |
 | --- | --- | --- | --- | --- |
 | matchDomains | Yes | An array of DNS zones or subdomains that can be accessed over this proxy | Array of Strings | [ "example.com" ] |
-| excludeDomains | Yes | An array of DNS zones or subdomains that cannot be accessed over this proxy. If matchDomains is specfied excludeDomains should be more specific domains of entries in the matchDomains array | Array of Strings | [ "public.example.com" ] |
+| excludeDomains | Yes | An array of DNS zones or subdomains that cannot be accessed over this proxy. If matchDomains is specfied, excludeDomains should list more specific domains within entries in the matchDomains array | Array of Strings | [ "public.example.com" ] |
 
 When present in a PvD Additional Information dictionary that is retrieved for a proxy
 as described in {{proxy-pvd}}, domains in the `matchDomains` array indicate specific zones
 that are accessible using the proxy. If a hostname is not included in the enumerated
 zones, then a client SHOULD assume that the hostname will not be accessible through the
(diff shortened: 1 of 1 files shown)


draft-ietf-intarea-proxy-config.md
```diff
@@ -249,14 +249,16 @@ array that are used for split-DNS configuration.
 | JSON Key | Optional | Description | Type | Example |
 | --- | --- | --- | --- | --- |
 | matchDomains | Yes | An array of DNS zones or subdomains that can be accessed over this proxy | Array of Strings | [ "example.com" ] |
-| excludedDomains | Yes | An array of DNS zones or subdomains that cannot be accessed over this proxy, which should be more specific domains of entries in the matchDomains array | Array of Strings | [ "public.example.com" ] |
+| excludeDomains | Yes | An array of DNS zones or subdomains that cannot be accessed over this proxy. If matchDomains is specfied excludeDomains should be more specific domains of entries in the matchDomains array | Array of Strings | [ "public.example.com" ] |
```
tfpauly, line 252:
```suggestion
| excludeDomains | Yes | An array of DNS zones or subdomains that cannot be accessed over this proxy. If matchDomains is specfied, excludeDomains should list more specific domains within entries in the matchDomains array | Array of Strings | [ "public.example.com" ] |
```

draft-ietf-intarea-proxy-config.md
```diff
@@ -249,14 +249,16 @@ array that are used for split-DNS configuration.
...
 
 When present in a PvD Additional Information dictionary that is retrieved for a proxy
 as described in {{proxy-pvd}}, domains in the `matchDomains` array indicate specific zones
 that are accessible using the proxy. If a hostname is not included in the enumerated
 zones, then a client SHOULD assume that the hostname will not be accessible through the
-proxy. If a hostname is included in the `excludedDomains` array, then the client SHOULD NOT
-access it through the proxy.
+proxy. If a hostname is included in the `excludeDomains` array, then the client SHOULD NOT
+access it through the proxy. The excludeDomains parameter can be present even if matchDomains
```
tfpauly, line 259:
```suggestion
access it through the proxy. The `excludeDomains` parameter can be present even if `matchDomains`
```

draft-ietf-intarea-proxy-config.md
```diff
@@ -249,14 +249,16 @@ array that are used for split-DNS configuration.
...
+is omitted. That means the client should assume that all domains except the domains
```
tfpauly, line 260:
```suggestion
is omitted. When this is the case, the client assumes that all domains except the domains
```

tfpauly (approved):

PR #269, opened 2024-12-01
Port matching rework

yaroslavros:
As discussed at IETF121 I merged matchPorts into other matching criteria. Adjusted examples and wording accordingly.

### DIFF
--- a/draft-ietf-intarea-proxy-config.md
+++ b/draft-ietf-intarea-proxy-config.md
@@ -265,46 +265,46 @@ be empty and MUST contain at least one valid key. Each provided key MUST corresp
 | ports | An array of TCP and UDP port ranges | Array of Strings | ["80", "443", "1024-65535"] |
 
 When present in a PvD Additional Information dictionary that is retrieved for a proxy
-as described in {{proxy-pvd}}, entries in the `domains` array of `match` object indicate
+as described in {{proxy-pvd}}, entries in the `domains` array of a `match` object indicate
 specific FQDNs and zones that are accessible using the proxy. If a hostname does not match any entry,
 then a client SHOULD assume that the hostname will not be accessible through the proxy.
-If a hostname is included in the `domains` array of `exclude` object, then the client SHOULD NOT
-access it through the proxy. The `domains` array can be present in `exclude` object even if `domains`
-is omitted from `match` object or `match` key is not present in the `proxies` array subdictionary.
+If a hostname is included in the `domains` array of an `exclude` object, then the client SHOULD NOT
+access it through the proxy. The `domains` array can be present in an `exclude` object even if `domains`
+is omitted from all `match` objects or `match` key is not present in the `proxies` array subdictionary.
 When this is the case, the client assumes that all domains except the domains
-listed in the `domains` from `exclude` object are accessible through the proxy. If `domains` arrays are
-specified in both `match` and `exclude` objects, the `domains` array from `exclude` object SHOULD
-list more specific domains within entries in the `domains` array from `match` object.
+listed in the `domains` from `exclude` objects are accessible through the proxy. If `domains` arrays are
+specified in `match` and `exclude` objects, `domains` arrays from `exclude` objects SHOULD
+list more specific domains within entries in `domains` arrays from `match` objects.
 
 A wildcard prefix (`*.`) is used to indicate matching entire domains or subdomains instead of specific hostnames. Note
 that this can be used to match multiple levels of subdomains. For example "*.example.com"
 matches "internal.example.com" as well as "www.public.example.com".
 Entries that include the wildcard prefix also SHOULD be treated as if they match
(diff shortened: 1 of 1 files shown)


draft-ietf-intarea-proxy-config.md
```diff
@@ -252,11 +252,10 @@ array that are used to signal information about destinations available through t
...
+| excludeDomains | Yes | An array of FQDNs and wildcard DNS domains (optionally with port specifications) that cannot be accessed over this proxy. If matchDomains is specified, excludeDomains should list more specific domains within entries in the matchDomains array | Array of Strings | [ "public.example.com:80" ] |
+| matchSubnets | Yes | An array of IP addresses and subnets (optionally with port specifications) accessible over this proxy | Array of Strings | [ "[2001:DB8::1]:443,1024-65535", "192.168.1.0/24" ] |
+| excludeSubnets | Yes | An array of IP addresses and subnets (optionally with port specifications) that cannot be accessed over this proxy. If matchSubnets is specified, excludeDomains should list more specific subnets within entries in the matchSubnets array | Array of Strings | [ "192.0.2.0/16:80", "192.51.100.1:1024-2048" ] |
 
 When present in a PvD Additional Information dictionary that is retrieved for a proxy
 as described in {{proxy-pvd}}, entries in the `matchDomains` array indicate specific FQDNs
```
ddragana, line 261: I think we need to define "optionally with port specification" before this.

Also FQDNs is not completely correct, because it can have a port specification. Probably we need to define matching algorithm, with and without port specification first.

draft-ietf-intarea-proxy-config.md
```diff
@@ -252,11 +252,10 @@ array that are used to signal information about destinations available through t
...
-| excludeDomains | Yes | An array of FQDNs and wildcard DNS domains that cannot be accessed over this proxy. If matchDomains is specified, excludeDomains should list more specific domains within entries in the matchDomains array | Array of Strings | [ "public.example.com" ] |
-| matchSubnets | Yes | An array of IP addresses and subnets that can be accessed over this proxy | Array of Strings | [ "2001:DB8::1", "192.168.1.0/24" ] |
-| excludeSubnets | Yes | An array of IP addresses and subnets that cannot be accessed over this proxy. If matchSubnets is specified, excludeDomains should list more specific subnets within entries in the matchSubnets array | Array of Strings | [ "192.0.2.0/16", "192.51.100.1" ] |
-| matchPorts | Yes | An array of TCP or UDP port ranges accessible over this proxy | Array of Strings | [ "80", "443", "1024-65535" ] |
+| matchDomains | Yes | An array of FQDNs and wildcard DNS domains (optionally with port specifications) accessible over this proxy | Array of Strings | [ "www.example.com:80,443", "*.internal.example.com" ] |
```
ddragana, line 255: > "www.example.com:80,443"

I am not sure about the syntax.

What do you thing to change to array of pairs, e.g:
[["www.example.com", "80,443"], ["public.example.com"]]

yaroslavros: Frankly I'm a bit torn. Proprietary steering configurations that I worked with previously always used a simple string with an optional colon to separate ports section. Also mandating nested lists for those who don't need ports feels like an over-engineering and likely to be error-prone. On another hand structure is almost always a good thing.

ddragana: I was look from perspective of parsing. More structure is easier to parse.

yaroslavros: List of lists with one or two elements seem quite ugly and unintuitive to me.
How about merging matchDomains and matchSubnets into a single "match" list made of dictionaries? Each dictionary may have a destination key matching criteria, which could be "ipv4", "ipv6" or "domain" and optionally "ports". IP address matching would cover specific addresses as well as subnets.
Similar logic could apply to "exclude".

ddragana: You are right that list of lists are not great.

I like your proposal.

yaroslavros: Thank you, @ddragana

@tfpauly do you agree with the proposed format?

draft-ietf-intarea-proxy-config.md
yaroslavros: I have updated description the text in description column and changed "(optionally with port specifications)" to "May include port specification" at the end of each description.
Port specification is defined under the table after all the keys.

What do you think?

yaroslavros:
@ddragana, I updated the PR to reflect our previous discussion on object based formatting for matching and exclusion criteria. Please review. Thank you!

yaroslavros:
I just realised that the new format takes us back to the problem that this PR was designed to solve in the first place!
Making one more change...

yaroslavros:
Another correction to clarify the possibility of multiple match and exclude objects. Also fixed the example.

tfpauly (approved):

PR #272, opened 2025-02-28
Added mandatory key

yaroslavros:
This makes format extensible in the future and allows proprietary extensions.
On top of that it addresses the question of what clients should do if they cannot fully process proxy destination restrictions. If corresponding key is in `mandatory` list, they must ignore the proxy, otherwise they may proceed even without full processing of the destination restriction list.

### DIFF
--- a/draft-ietf-intarea-proxy-config.md
+++ b/draft-ietf-intarea-proxy-config.md
@@ -160,7 +160,7 @@ use of a proxy.
 
 Optional keys included in this or future specifications MAY be included
 in the `mandatory` list. Clients that do not recognize a key from the `mandatory`
-list or fully process corresponding sub-dictionary entry MUST ignore the entire proxy
+list or cannot fully process the corresponding sub-dictionary entry MUST ignore the entire proxy
 definition, since the proxy might be only applicable for particular uses. These keys are
 registered in an IANA registry, defined in {{proxy-info-iana}}. `mandatory` list MUST NOT
 include any entries that do not correspond to optional keys in the sub-dictionary.
--- a/draft-ietf-intarea-proxy-config.md
+++ b/draft-ietf-intarea-proxy-config.md
@@ -165,10 +165,10 @@ use of a proxy.
 | alpn | Yes | An array of Application-Layer Protocol Negotiation protocol identifiers | Array of Strings | ["h3","h2"] |
 | mandatory | Yes | An array of optional keys that client must understand and process to use this proxy | Array of Strings | ["matchDomains"] |
 
-The value of the `mandatory` key is a list of keys that the client must understand and process to be 
-able to use the proxy. A client that does not understand a key from the list or cannot fully process 
-the value of a key from the list MUST ignore the entire proxy definition. The list can contain 
-only keys that are registered in an IANA registry, defined in {{proxy-info-iana}} and that are marked 
+The value of the `mandatory` key is a list of keys that the client must understand and process to be
+able to use the proxy. A client that does not understand a key from the list or cannot fully process
+the value of a key from the list MUST ignore the entire proxy definition. The list can contain
+only keys that are registered in an IANA registry, defined in {{proxy-info-iana}} and that are marked
 as optional.  The `mandatory` list MUST NOT include any entries that are not present in the sub-dictionary.
 
 The values for the `protocol` key are defined in the proxy protocol
--- a/draft-ietf-intarea-proxy-config.md
+++ b/draft-ietf-intarea-proxy-config.md
@@ -154,18 +154,23 @@ Each proxy is defined by a proxy protocol, a proxy location (i.e., a hostname an
 
 This document defines two mandatory keys for the sub-dictionaries in the
 `proxies` array, `protocol` and `proxy`. There are also optional keys, including
-`alpn`, and destination accessibility keys defined in {{destinations}}.
-Other optional keys can be added to the dictionary
-to further define or restrict the use of a proxy. Clients that do not
-recognize or understand a key in a proxy sub-dictionary MUST ignore the entire
-proxy definition, since the proxy might be only applicable for particular
-uses. These keys are registered in an IANA registry, defined in {{proxy-info-iana}}.
+`alpn`, `mandatory`, and destination accessibility keys defined in {{destinations}}.
+Other optional keys can be added to the dictionary to further define or restrict the
+use of a proxy.
+
+Optional keys included in this or future specifications MAY be included
+in the `mandatory` list. Clients that do not recognize a key from the `mandatory`
+list or fully process corresponding sub-dictionary entry MUST ignore the entire proxy
+definition, since the proxy might be only applicable for particular uses. These keys are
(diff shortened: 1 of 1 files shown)


draft-ietf-intarea-proxy-config.md
```diff
@@ -154,18 +154,23 @@ Each proxy is defined by a proxy protocol, a proxy location (i.e., a hostname an
...
-uses. These keys are registered in an IANA registry, defined in {{proxy-info-iana}}.
+`alpn`, `mandatory`, and destination accessibility keys defined in {{destinations}}.
+Other optional keys can be added to the dictionary to further define or restrict the
+use of a proxy.
+
+Optional keys included in this or future specifications MAY be included
+in the `mandatory` list. Clients that do not recognize a key from the `mandatory`
+list or fully process corresponding sub-dictionary entry MUST ignore the entire proxy
+definition, since the proxy might be only applicable for particular uses. These keys are
+registered in an IANA registry, defined in {{proxy-info-iana}}. `mandatory` list MUST NOT
```
ddragana, line 165: what about proprietary keys? I think mandatory list can contain any key except the one that define as mandatory in IANA registry.

yaroslavros: I'm planning to propose proprietary keys (as well as ssh-like key namespaces) in a separate PR.
Is that okay with you?

draft-ietf-intarea-proxy-config.md
```diff
@@ -154,18 +154,23 @@ Each proxy is defined by a proxy protocol, a proxy location (i.e., a hostname an
...
+`alpn`, `mandatory`, and destination accessibility keys defined in {{destinations}}.
+Other optional keys can be added to the dictionary to further define or restrict the
+use of a proxy.
+
+Optional keys included in this or future specifications MAY be included
+in the `mandatory` list. Clients that do not recognize a key from the `mandatory`
+list or cannot fully process the corresponding sub-dictionary entry MUST ignore the entire proxy
+definition, since the proxy might be only applicable for particular uses. These keys are
+registered in an IANA registry, defined in {{proxy-info-iana}}. `mandatory` list MUST NOT
+include any entries that do not correspond to optional keys in the sub-dictionary.
```
ddragana, line 166: we should move this below the table. The keys in the table are all define below.

yaroslavros: Done

draft-ietf-intarea-proxy-config.md
```diff
@@ -154,18 +154,23 @@ Each proxy is defined by a proxy protocol, a proxy location (i.e., a hostname an
...
 | proxy | No | String containing the URI template or hostname and port of the proxy, depending on the format defined by the protocol | String | "https://proxy.example.org:4443/masque{?target_host,target_port}" |
 | alpn | Yes | An array of Application-Layer Protocol Negotiation protocol identifiers | Array of Strings | ["h3","h2"] |
+| mandatory | Yes | An array of optional keys that client must understand and process to use this proxy | Array of Strings | ["matchDomains"] |
+
+Optional keys included in this or future specifications MAY be included
+in the `mandatory` list. Clients that do not recognize a key from the `mandatory`
+list or cannot fully process the corresponding sub-dictionary entry MUST ignore the entire proxy
+definition, since the proxy might be only applicable for particular uses. These keys are
+registered in an IANA registry, defined in {{proxy-info-iana}}. `mandatory` list MUST NOT
+include any entries that do not correspond to optional keys in the sub-dictionary.
```
ddragana, line 173:
```suggestion
The value of the "mandatory" key is a list of key that the client must understand and process to be 
able to use the proxy. A client that does not understand a key from the list or cannot fully process 
the value of a key from the list MUST ignore the entire proxy definition. The list can contain 
only keys that are registered in an IANA registry, defined in {{proxy-info-iana}} and that are marked 
as optional.  The `mandatory` list MUST NOT include any entries that are not present in the sub-dictionary.
```

yaroslavros: Rephrased as requested. Corrected few minor mistakes in the suggestion.

yaroslavros:
@ddragana I believe I resolved all the issues you raised. You okay to merge this?

draft-ietf-intarea-proxy-config.md
```diff
@@ -154,18 +154,22 @@ Each proxy is defined by a proxy protocol, a proxy location (i.e., a hostname an
...
 | protocol | No | The protocol used to communicate with the proxy | String | "connect-udp" |
 | proxy | No | String containing the URI template or hostname and port of the proxy, depending on the format defined by the protocol | String | "https://proxy.example.org:4443/masque{?target_host,target_port}" |
 | alpn | Yes | An array of Application-Layer Protocol Negotiation protocol identifiers | Array of Strings | ["h3","h2"] |
+| mandatory | Yes | An array of optional keys that client must understand and process to use this proxy | Array of Strings | ["matchDomains"] |
+
+The value of the `mandatory` key is a list of keys that the client must understand and process to be
+able to use the proxy. A client that does not understand a key from the list or cannot fully process
+the value of a key from the list MUST ignore the entire proxy definition. The list can contain
+only keys that are registered in an IANA registry, defined in {{proxy-info-iana}} and that are marked
+as optional.  The `mandatory` list MUST NOT include any entries that are not present in the sub-dictionary.
```
ddragana, line 172: Sorry, this is on me. I was not precise enough.

Can you move this below alpn, so that the keys are discussed in the same order as in the table.
Thanks!

draft-ietf-intarea-proxy-config.md
```diff
@@ -154,18 +154,22 @@ Each proxy is defined by a proxy protocol, a proxy location (i.e., a hostname an
...
 | --- | --- | --- | --- | --- |
 | protocol | No | The protocol used to communicate with the proxy | String | "connect-udp" |
 | proxy | No | String containing the URI template or hostname and port of the proxy, depending on the format defined by the protocol | String | "https://proxy.example.org:4443/masque{?target_host,target_port}" |
 | alpn | Yes | An array of Application-Layer Protocol Negotiation protocol identifiers | Array of Strings | ["h3","h2"] |
+| mandatory | Yes | An array of optional keys that client must understand and process to use this proxy | Array of Strings | ["matchDomains"] |
+
+The value of the `mandatory` key is a list of keys that the client must understand and process to be
+able to use the proxy. A client that does not understand a key from the list or cannot fully process
+the value of a key from the list MUST ignore the entire proxy definition. The list can contain
+only keys that are registered in an IANA registry, defined in {{proxy-info-iana}} and that are marked
```
ddragana, line 171: you are going to change this in the pr for proprietary keys?
I  am wondering if we should change it in this pr because the other may not be ready before the deadline today.

yaroslavros: I would merge this one as it is and add proprietary keys in a separate PR that we don't need to rush prior to the deadline tonight.
Proprietary keys format will depend on allowed characters in the JSON keys and I don't think we have full consensus on that.

tfpauly (approved):

PR #279, opened 2025-05-11
Versioning support

yaroslavros:
Also introduces a simple mechanism for client to discover new configuration version from proxy responses

### DIFF
--- a/draft-ietf-intarea-proxy-config.md
+++ b/draft-ietf-intarea-proxy-config.md
@@ -362,6 +362,34 @@ content-length = 135
 The client could then choose to use this proxy only for accessing names that fall
 within the "*.internal.example.org" zone.
 
+# Configuration Versioning {#config-versioning}
+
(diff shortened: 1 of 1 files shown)


tfpauly:
A couple notes here:
- "version" in the name is confusing, since it can sound like a protocol version or configuration structure version. Naming would be good to fix.
- There could be other spellings, like the client providing the `etag` of the PvD it had used, and the proxy telling the client that its PvD is out of date

tfpauly:
This may also be something we could split out into a different version

tfpauly:
There is also a problem of a proxy being used by multiple configurations, and it wouldn't know which one the client was using to talk to it.

tfpauly:
Cases:
- This proxy no longer supports this target (hard failure)
- This proxy will accept this for now, but the config seems old (soft failure)
- There are new targets that the client may not be proxying, needs to expand and update

PR #280, opened 2025-05-11
Moved individual proxy protocols to informative references

yaroslavros:
I don't think implementers of this spec need to know intimate details of CONNECT-IP, CONNECT-UDP, CONNECT-TCP or SOCKSv5.


tfpauly (commented):
I'm not sure I agree with this — while you don't need to read those documents to be able to implement this spec, those documents do define the URI templates they use normatively...

yaroslavros:
I don't have a strong opinion on this one, but thought it would be a good idea not to be blocked by templated CONNECT-TCP.
Let's leave it as it is for now.

PR #281, opened 2025-05-20
Match/exclude destination rules (flattened)

yaroslavros:
I have collapsed all my suggestions from https://github.com/tfpauly/privacy-proxy/pull/276 into a fresh PR so it could be readable.

PR #283, opened 2025-06-16
Editorial pass

tfpauly:
Editorial pass after merging #281

### DIFF
--- a/draft-ietf-intarea-proxy-config.md
+++ b/draft-ietf-intarea-proxy-config.md
@@ -32,7 +32,7 @@ author:
 
 This document defines a mechanism for accessing provisioning domain information
 associated with a proxy, such as other proxy URIs that support different protocols
-and a list of DNS zones that are accessible via a proxy.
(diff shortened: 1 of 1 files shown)


draft-ietf-intarea-proxy-config.md
```diff
@@ -433,6 +485,12 @@ Further security and experience considerations are needed for these cases.
 
 # Security Considerations {#sec-considerations}
 
+The mechanisms in this document allow clients using a proxy to "upgrade" a configuration
+for an insecure proxy into a configuration for using TLS to communication to the proxy.
```
yaroslavros, line 489: I think "insecure proxy" term needs to be defined or replaced with something more precise.
I suspect you are referring to cleartext HTTP/1 proxy?

tfpauly:
```suggestion
for a cleartext HTTP/1.1 or SOCKS proxy into a configuration that uses TLS to communication to the proxy.
```

PR #295, opened 2025-07-22
Allow IP subnet rules to apply to resolved names

### LINKED ISSUE #282: Allow IP-based rules after a DNS lookup
"Subnet-based destination information are only meant be used when applications are communicating with destinations identified by only an IP address, and not a hostname.
Destination rules SHOULD NOT contain both the domains key and the subnets key. Many clients will not resolve a domain being accessed through the proxy to an IP address before using a proxy, so the subnet information may not be available."

It is very common for almost every PAC file in use today to have lines that send traffic to internal subnets DIRECT, based on a DNS lookup.  This is due to the common practice of internally hosted and externally hosted applications using the same domain.  The list to bypass these internally hosted sites one at a time, along with the maintenance of that list would be a large undertaking at any decently sized organization.  

I understand the desire to prevent DNS requests by the host, but this would prevent the uptake of this standard in large corporate environments.

tfpauly:
Closes #282

### DIFF
--- a/draft-ietf-intarea-proxy-config.md
+++ b/draft-ietf-intarea-proxy-config.md
@@ -357,12 +357,12 @@ This is done to prevent commonly needing to include both "\*.example.com" and "e
 in the `domains` array of a `proxy-match` rule.
 
 The `subnets` array includes IPv4 and IPv6 address literals, as well as IPv4 and IPv6 address subnets
-written using CIDR notation. Subnet-based destination information is only meant be used when applications
-are communicating with destinations identified by only an IP address, and not a hostname.
-
-Destination rules SHOULD NOT contain both the `domains` key and the `subnets` key. Many
-clients will not resolve a domain being accessed through the proxy to an IP address before using
-a proxy, so the subnet information may not be available.
+written using CIDR notation. Subnet-based destination information can apply to cases where
+applications are communicating directly with an IP address (without having resolved a DNS name)
+as well as cases where an application resolved a DNS name to a set of IP addresses. Note that
+if destination rules includes an empty `proxies` list (indicating that no proxy is applicable for
+this subnet), an application can only reliably follow this destination rule if it resolves DNS
+names prior to proxying.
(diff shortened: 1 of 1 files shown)


draft-ietf-intarea-proxy-config.md
```diff
@@ -357,12 +357,12 @@ This is done to prevent commonly needing to include both "\*.example.com" and "e
...
 The `subnets` array includes IPv4 and IPv6 address literals, as well as IPv4 and IPv6 address subnets
-written using CIDR notation. Subnet-based destination information is only meant be used when applications
-are communicating with destinations identified by only an IP address, and not a hostname.
-
-Destination rules SHOULD NOT contain both the `domains` key and the `subnets` key. Many
-clients will not resolve a domain being accessed through the proxy to an IP address before using
-a proxy, so the subnet information may not be available.
+written using CIDR notation. Subnet-based destination information can apply to cases where
+applications are communicating directly with an IP address (without having resolved a DNS name)
+as well as cases where an application resolved a DNS name to a set of IP addresses. Note that
```
ddragana, line 362: Do we care if the all IP from the resolved set are in the subnets or just a subset?

draft-ietf-intarea-proxy-config.md
```diff
@@ -357,12 +357,12 @@ This is done to prevent commonly needing to include both "\*.example.com" and "e
 in the `domains` array of a `proxy-match` rule.
 
 The `subnets` array includes IPv4 and IPv6 address literals, as well as IPv4 and IPv6 address subnets
-written using CIDR notation. Subnet-based destination information is only meant be used when applications
-are communicating with destinations identified by only an IP address, and not a hostname.
-
-Destination rules SHOULD NOT contain both the `domains` key and the `subnets` key. Many
```
ddragana, line 363: Was removing SHOULD NOT on purpose?

ddragana (approved):

PR #316, opened 2025-11-06
Clarify characters in identifier string

### LINKED ISSUE #310: Clarify characters in "identifier" string
From David Schinazi

> * what characters are allowed in the "identifier" string? Is it encoded as
> ASCII or UTF-8?

tfpauly:
Closes #310 
Clarify the format of the identifier value in proxy dictionaries.

### DIFF
--- a/draft-ietf-intarea-proxy-config.md
+++ b/draft-ietf-intarea-proxy-config.md
@@ -254,8 +254,8 @@ this can indicate if the proxy supports HTTP/3, HTTP/2, etc.
 
 The value of `identifier` key is a string that can be used to refer to a particular
 proxy from other dictionaries, specifically those defined in {{destinations}}. The
-string value is an arbitrary JSON string (allowing any format valid for a JSON string
-as defined in {{Section 7 of JSON}}). Identifier values MAY be duplicated
+string value is an arbitrary non-empty JSON string using UTF-8 encoding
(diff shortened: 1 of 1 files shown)


draft-ietf-intarea-proxy-config.md
```diff
@@ -254,7 +254,8 @@ this can indicate if the proxy supports HTTP/3, HTTP/2, etc.
 
 The value of `identifier` key is a string that can be used to refer to a particular
 proxy from other dictionaries, specifically those defined in {{destinations}}. The
-string value is an arbitrary JSON string. Identifier values MAY be duplicated
+string value is an arbitrary JSON string (allowing any format valid for a JSON string
+as defined in {{Section 7 of JSON}}). Identifier values MAY be duplicated
```
yaroslavros: Maybe we should be explicit that any UTF-8 string is valid as long as escape rule are followed?
Also I think it's a good idea to mention that the string is not empty.

Something along the lines of
```suggestion
string value is an arbitrary non-empty JSON string. According to {{Section 8.1 of JSON}} UTF-8 characters are allowed as
long as they are correctly represented as defined in {{Section 7 of JSON}}). Identifier values MAY be duplicated
```

What do you think?

yaroslavros (approved):

PR #334, opened 2026-03-06
Clarification on PvD URI usage

tfpauly:
Address David's comments. Clarify that the fetch is of a URI, and that URI has a default derivation (well-known using the hostname) but can be anything else if configured.

### DIFF
--- a/draft-ietf-intarea-proxy-config.md
+++ b/draft-ietf-intarea-proxy-config.md
@@ -126,7 +126,7 @@ type as defined in {{Section 4.1 of PVDDATA}}. The fetch MUST use the "https" sc
 
 {{PVDDATA}} defines the well-known PvD URI, ".well-known/pvd", that is served on the
 stardard port for HTTP over TLS (HTTPS), port 443. When a client is provisioned
-with the hostname of a proxy (such as a SOCKS proxy or an HTTP CONNECT proxy) for
+with the hostname of a proxy for
 which it wants to look up PvD Additional Information, the client SHALL use the
 well-known PvD URI using the host authority of the proxy. A client can also be directly
 configured with a HTTPS URI on which to fetch the PvD Information, in which case the


draft-ietf-intarea-proxy-config.md
```diff
@@ -120,13 +120,17 @@ inconsistencies and can open up security vulnerabilities.
...
+an HTTP GET request for a PvD URI using the "application/pvd+json" media
+type as defined in {{Section 4.1 of PVDDATA}}. The fetch MUST use the "https" scheme.
+
+{{PVDDATA}} defines the well-known PvD URI, ".well-known/pvd", that is served on the
+stardard port for HTTP over TLS (HTTPS), port 443. When a client is provisioned
+with the hostname of a proxy (such as a SOCKS proxy or an HTTP CONNECT proxy) for
+which it wants to look up PvD Additional Information, the client SHALL use the
+well-known PvD URI using the host authority of the proxy. A client can also be directly
+configured with a HTTPS URI on which to fetch the PvD Information, in which case the
+fetch SHALL be made to that configured URI.
```
yaroslavros: I think technically "provisioned with the hostname of a proxy" applies to everything as hostname is a part of URI.

```suggestion
{{PVDDATA}} defines the well-known PvD URI, ".well-known/pvd", that is served on the
stardard port for HTTP over TLS (HTTPS), port 443. When a client is provisioned
with a proxy defined by a hostname rather than full URI (such as a SOCKS proxy or an HTTP CONNECT proxy) for
which it wants to look up PvD Additional Information, the client SHALL use the
well-known PvD URI using the host authority of the proxy. A client can also be directly
configured with a HTTPS URI on which to fetch the PvD Information, in which case the
fetch SHALL be made to that configured URI.
```

tfpauly: I want to make sure it's clear that when you have a masque-style URI, you still look up the PvD based on the hostname (unless you know a full URL some other way)


```suggestion
{{PVDDATA}} defines the well-known PvD URI, ".well-known/pvd", that is served on the
stardard port for HTTP over TLS (HTTPS), port 443. When a client is provisioned
with the hostname of a proxy (such as for a SOCKS proxy or an HTTP CONNECT proxy, or as part of a complete proxy URI) for
which it wants to look up PvD Additional Information, the client SHALL use the
well-known PvD URI using the host authority of the proxy. A client can also be directly
configured with a HTTPS URI on which to fetch the PvD Information, in which case the
fetch SHALL be made to that configured URI.
```

yaroslavros (approved):

PR #354, opened 2026-05-07
be explicit this is not a generic proxy discovery mechanism + IPv6

boucadair:


draft-ietf-intarea-proxy-config.md
```diff
@@ -40,7 +40,8 @@ informative:
 
 This document defines a mechanism for accessing provisioning domain information
 associated with a proxy, such as other proxy URIs that support different protocols
-and information about which destinations are accessible using a proxy.
+and information about which destinations are accessible using a proxy. Note that this
+mechanism is not intended to be used as a generic method for any type of proxies (SIP, etc.).
```
tfpauly, line 44: I don't think this is correct to state. SIP proxies are configurable via URLs (see [example](https://www.cisco.com/c/en/us/td/docs/voice_ip_comm/cust_contact/contact_center/customer_voice_portal/cvp_12_5/configuration/guide/ccvp_b_configuration-guide-12-5-1/ccvp_b_configuration-guide-12-5-1_chapter_01110.pdf)).

SIP proxies could be easily added to Table 2 (Initial PvD Proxy Protocol Registry Contents).

I'm not aware of any types of proxies that would not be appropriate to advertise in this mechanism.

draft-ietf-intarea-proxy-config.md
```diff
@@ -99,7 +100,8 @@ a full URI template {{URITEMPLATE}}.
 The mechanisms defined in this document are intended to offer a standard
 alternative that works for URI-based proxies and avoids dependencies
 on executing JavaScript scripts, which are prone to implementation inconsistencies
-and security vulnerabilities.
+and security vulnerabilities. However, the mechanism is applicable only when an IPv6
+connectivity is provided.
```
tfpauly, line 104: This is not correct — nothing in sections 2-4 has any requirement for clients to have IPv6 connectivity. In section 5, the "out of scope" mechanism we refer to does require being able to receive RAs, but that doesn't require non-link-local connectivity over IPv6 to work.

tfpauly (changes requested):

draft-ietf-intarea-proxy-config.md
boucadair: There are already existing mechanisms to configure such proxies. I'm concerned that we are adding yet another mechanisms to discover these services. If we focus on SIP, do we have an operational problem todat to provision these services? What about conflict of various discovery channels, etc.

The pb you had in the draft (PAC, limitations etc.) was reasonable as it is seemed to me that it targeted a very specific set of proxies.

draft-ietf-intarea-proxy-config.md
boucadair: I'm may be missing something, but I understood PvD Additional Information as defined in rfc8801#section-4. I need to recheck the text.

draft-ietf-intarea-proxy-config.md
tfpauly: I'm not suggesting that people use this instead of other mechanisms specifically for SIP, but there is no inherent limitation here that would make this mechanism inappropriate for SIP discovery. There's not a limited set of proxies.

draft-ietf-intarea-proxy-config.md
tfpauly: Yeah, the PvD Additional Information is not specific to IPv6.

evyncke: I can only second Tommy's point:
> hat doesn't require non-link-local connectivity over IPv6 to work.

